PR #13, opened 2018-09-08
iOS & Android integration tests

radex:
Changed:

- Copied over / set up again the integration test harness from Purple
- Added missing Metro/Babel/ESLint configuration + dependencies necessary to package watermelon through metro for integration tests and for all checks to pass
- Added `yarn test:ios` and `yarn test:android`
- Hooked up test:ios into Travis CI
- Moved BridgeTestReporter from the main Watermelon bundle to the integration test harness apps (🍉 clients don't need that)
- Moved `__tests__` inside src
- Wrote up contribution guidelines for native code

Known issues:

- Android tests don't run on CI yet
- I couldn't figure out how to make Metro understand our absolute paths without setting up rn-cli.js, transformer, and a bunch of package.jsons (Metro finds those files using Haste). I'll make a separate issue for this

PR #14, opened 2018-09-08
Add quotes to names in sql queries to allow keywords as table names and columns

### LINKED ISSUE #9: Column names not escaped properly
When using SQL keywords as column names, the application dies on launch.

Example:
```
tableSchema({
      name: 'ticket_statuses',
      columns: [
        { name: 'name', type: 'string' },
        { name: 'order', type: 'string' },
      ]
    })
```

while *order* is a SQL keyword. Stacktrace shows `Syntax error near "order"`

sebastian-schlecht:
@radex gave this a quick shot. Is there any way to integration-test this out-of-the-box? 

Fixes #9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+
+### Fixed
 - Add quotes to all names in sql queries to allow keywords as table or column names
 
 ## [0.6.0] - 2018-09-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-### Addded
+### Added
 
 ### Changed
+- Add quotes to all names in sql queries to allow keywords as table or column names
 
 ## [0.6.0] - 2018-09-05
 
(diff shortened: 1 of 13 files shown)


radex:
Good work! Thank you.

Here's a good way to add an integration test: https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/__tests__/commonTests.js#L360

I'd add a new table with keywords as columns to the test schema: https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/__tests__/helpers.js

and then in the test you can test the basic functionality: insert new item to that particular table, query it, update it, delete it permanently, mark as deleted, remove deleted records. See the tests above in the commonTests file for examples of the adapter api.

Then to test it, this doesn't work on master yet, but if you check your changes on this branch: https://github.com/Nozbe/WatermelonDB/pull/13 you can do `yarn test:ios` and/or `yarn test:android` to run integration tests (run the code through SQLite).

Let me know if you run into any issues

sebastian-schlecht:
Cool, I will fiddle with that, thanks.

sebastian-schlecht:
Alright, added keywords to the common tests and to the test schema. Tests pass on your branch.

radex:
Cool, I think this is good enough — although you could make the tests better by adding (or extending) a test to make sure that querying with a keyword like `order` actually works. What you had added ensures that schema, insert, and maybe update with this work, but I don't see a test that puts a keyword in query terms

radex:
LGTM. Please tweak the changelog and let me know if you want to play a bit more with the tests as suggested in the previous comment

radex:
Oops, actually it’s red. Run ‘yarn ci:check’ locally — there’s an eslint issue

sebastian-schlecht:
Oh my, missed that one - updated.

radex:
Woohoo 🎉 merged!

PR #18, opened 2018-09-09
Don't rely on Jest to mock lokijs web worker in tests

radex:
- This fixes Purple tests. Before, Jest magic was required to use a web worker mock in tests. Now, it's just NODE_ENV check and conditional require, so it works in Purple out of the box
- Improves build script (doesn't copy over unnecessary native files)

### DIFF
--- a/src/adapters/lokijs/WorkerBridge.js
+++ b/src/adapters/lokijs/WorkerBridge.js
@@ -25,8 +25,7 @@ class WorkerBridge {
   _pendingRequests: WorkerActions = []
 
   _createWorker(): Worker {
-    const WorkerClass = (LokiWorker: any)
-    const worker: Worker = new WorkerClass()
+    const worker: Worker = (new LokiWorker(): any)
 
     worker.onmessage = ({ data }) => {
       const { type, payload }: WorkerResponseAction = (data: any)
--- a/src/adapters/lokijs/WorkerBridge.js
+++ b/src/adapters/lokijs/WorkerBridge.js
@@ -25,7 +25,8 @@ class WorkerBridge {
   _pendingRequests: WorkerActions = []
 
   _createWorker(): Worker {
-    const worker: Worker = new LokiWorker()
+    const WorkerClass = (LokiWorker: any)
+    const worker: Worker = new WorkerClass()
 
     worker.onmessage = ({ data }) => {
       const { type, payload }: WorkerResponseAction = (data: any)
--- a/src/adapters/lokijs/worker/index.worker.js
+++ b/src/adapters/lokijs/worker/index.worker.js
@@ -5,18 +5,17 @@
 
 import LokiWorker from './lokiWorker'
 
-let defaultExport
-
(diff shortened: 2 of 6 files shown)


radex:
@kokusGr CR plz

src/adapters/lokijs/worker/index.worker.js
```diff
@@ -1,10 +1,22 @@
...
+if (process.env.NODE_ENV === 'test') {
+  const workerMock = require('./workerMock').default
+  defaultExport = workerMock
+} else {
   self.workerClass = new LokiWorker(self)
+  defaultExport = self
 }
 
-export default self
+export default defaultExport
```
kokusGr: I think we can avoid mutating variable and disabling eslint rule by creating a helper function. Something like this:
```
const getDefaultExport = () => {
  if (process.env.NODE_ENV === 'test') {
    const workerMock = require('./workerMock').default
    return workerMock
  }

  self.workerClass = new LokiWorker(self)
  return self
}
export default getDefaultExport()
```
WDYT?

src/adapters/lokijs/WorkerBridge.js
```diff
@@ -25,7 +25,8 @@ class WorkerBridge {
   _pendingRequests: WorkerActions = []
 
   _createWorker(): Worker {
-    const worker: Worker = new LokiWorker()
+    const WorkerClass = (LokiWorker: any)
+    const worker: Worker = new WorkerClass()
```
kokusGr: Why not `const worker: Worker = (new LokiWorker(): any)`?

radex: the problem is with `LokiWorker`, because Flow doesn't understand it's always a Class, and thinks it might be `any` (and therefore `new x()` is illegal)

src/adapters/lokijs/worker/index.worker.js
radex: 👍

src/adapters/lokijs/WorkerBridge.js
kokusGr: Hmm are you sure? I just changed it locally and `ci:check` shows no errors.

radex: I'll try again ;)

radex:
@kokusGr You're right, it works. Fixed 🏓

src/adapters/lokijs/worker/index.worker.js
```diff
@@ -1,10 +1,21 @@
 // @flow
 /* eslint-disable no-restricted-globals */
+/* eslint-disable global-require */
+/* eslint-disable import/no-mutable-exports */
```
kokusGr: Don't need this now

radex: oops - fixed

kokusGr:
Ok, thx :)

radex:
🎉

PR #19, opened 2018-09-10
Fix Flow in apps using WatermelonDB

radex:
- Instead of copying 🍉 source files to the compiled NPM dist/ folder directly, it puts them in dist/src
- … but first runs Babel through those files to rewrite import paths to relative paths, since we use absolute paths and 3rd party apps have no idea what they mean.
- Unfortunately there's a Babel bug and the original comments in files are written wrong (there's often an empty line between a comment and the next line of code), so `// $FlowFixMe` wouldn't work. So I rewrote those to use `(x : any)`. Also fixed other issues that produced problems when running Flow in Purple
- added docs to explain how to use Flow in apps with Watermelon

### DIFF
--- a/src/observation/encodeMatcher/index.js
+++ b/src/observation/encodeMatcher/index.js
@@ -72,6 +72,5 @@ export default function encodeMatcher<Element: Model>(query: QueryDescription): 
 
   invariant(!join.length, `Queries with joins can't be encoded into a matcher`)
 
-  // $FlowFixMe
-  return encodeConditions(where)
+  return (encodeConditions(where): any)
 }
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -71,12 +71,12 @@ module.exports = {
     },
     rewriteonly: {
       plugins: [
-        '@babel/plugin-syntax-flow',
         '@babel/plugin-syntax-class-properties',
         ['@babel/plugin-syntax-decorators', { legacy: true }],
-        '@babel/plugin-syntax-jsx',
+        // '@babel/plugin-syntax-jsx',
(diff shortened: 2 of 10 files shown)


radex:
⚠️ don't merge — dependent branch

radex:
@kokusGr CR please ;)

src/observation/encodeMatcher/index.js
```diff
@@ -72,6 +72,5 @@ export default function encodeMatcher<Element: Model>(query: QueryDescription):
 
   invariant(!join.length, `Queries with joins can't be encoded into a matcher`)
 
-  // $FlowFixMe
```
kokusGr: I think we should add `TODO` or some other mark in place of `$FlowFixMe` so we could later on check/fix those flow errors.

kokusGr:
@radex One small thing and LGTM :)

src/observation/encodeMatcher/index.js
radex: Hm, don't you think `: any` is essentially the same as `$FlowFixMe`?

kokusGr: Kinda, but there are situations that `:any` is used to "change" type like `const worker: Worker = (new LokiWorker: any)`.

radex: Right, but they're used interchangeably — in both cases we do this to lie to Flow because something is broken or we can't figure it out. We want to avoid `any` the same way as we avoid `$FlowFixMe`. I think the only practical difference is that `(: any)` applies to an expression, whereas $FlowFixMe applies to the whole line. No?

radex:
Thanks! 🎉

PR #26, opened 2018-09-11
tests: use Node.js 10 and properly install Yarn

DanielRuf:


DanielRuf:
Hm, it seems `yarn` was replaced with `npm install`. Is there any good reason to switch back to npm?

DanielRuf:
> Well… is there any specific reason why Travis should use yarn? I use and recommend yarn in development, but I figured it'll save time on CI to just use npm…

Yarn and npm still do different things when it comes to lockfiles and install commands. Also Yarn is still more stable and in many cases faster than npm.

We have a yarn.lock. Yarn uses this and does not automatically update the versions of dependencies like npm does. Also for npm to get faster builds, you need package-lock.json + `npm ci`.

DanielRuf:
As we have no package-lock.json but the CI uses npm, it makes sense to remove yarn.lock then as it would lock the versions of contributors but use different versions on CI.

radex:
> Yarn and npm still do different things when it comes to lockfiles and install commands.

Ah… So you're saying that if I do `npm install` on CI, then it will fetch the latest matching versions of all packages, not what they're supposed to be locked on.

That makes sense. I don't think installing yarn on CI is enough of a big deal to justify maintaining two different lock files. 👍

DanielRuf:
> Ah… So you're saying that if I do npm install on CI, then it will fetch the latest matching versions of all packages, not what they're supposed to be locked on.

Exactly, this is what npm does. A lock file just makes a snapshot of an install. You can easily test this with an older package-lock.json and then run npm install after removing node_modules. The lockfile will change.

radex:
Let's then go back to yarn — is this PR mergeable as is?

DanielRuf:
> is this PR mergeable as is?

Yes, this PR can be merged and seems to be ok so far.

radex:
Great! Thank you @DanielRuf!

PR #34, opened 2018-09-14
Loki worker queue

rkrajewski:
Removes `async` dependency, used for worker's queue, and replaces it with simple queue based on `rxjs`.


radex:
Not very Rx idiomatic, but good enough! LGTM, thanks @rkrajewski 🎉

PR #38, opened 2018-09-17
Web demo styles

panr:
This PR makes web demo more pleasurable to use 😎 

web-demo


examples/web/src/components/Root/index.js
```diff
@@ -7,11 +7,12 @@ import BlogList from 'components/BlogList'
 import Post from 'components/Post'
 import Blog from 'components/Blog'
 import ModerationQueue from 'components/ModerationQueue'
+import logoSrc from './WatermelonLogo.svg'
 
 import style from './style'
 
 class Root extends Component {
-  state = { isGenerating: false }
+  state = { isGenerating: false, isGenerated: false }
```
radex: nope, `isGenerated: false` is not necessary, and breaks the app after you refresh. Look at examples/native. We just need to stop rendering the lists during generation

radex:
Looks great @panr, except for a bug when you refresh the app.

@kokusGr / @rkrajewski — want to CR?

kokusGr:
LGTM 👌

radex:
@panr I tweaked it a bit because we shouldn't observe the database at all when doing "generate x records" (clearing the whole database is currently an unsafe behavior and there shouldn't be any observers at that time). Look at my changes. Are they OK? If so, we can merge

panr:
> Are they OK? If so, we can merge

Looks fine 👍

PR #46, opened 2018-09-21
Relative imports

### LINKED ISSUE #20: [Discussion] Absolute vs relative paths
Currently, most imports in 🍉 are absolute (i.e. `import from 'Query'`). This is mostly an artifact of the fact that it was extracted from a codebase of an app. But it's also very convenient when working on Watermelon.

But it adds A LOT of complexity, because every tool that has to understand the codebase has to be (often separately) configured:

- Babel (for building) - #22
- Jest
- Eslint (via eslint-babel)
- Flow
- Metro (for integration tests)
- the app and Jest in an app importing Watermelon (we use Babel to rewrite for dist)
- Flow in the app importing Watermelon (also using Babel but have to write a separate copy with Flow annotations — which has its own problems… - see #21 )

I see 3 options:

1. We keep the current setup (This is a big PITA)
2. We give up and just change the absolute paths to relative paths.
3. We meet halfway and change all paths to something like `~/blah` or `src/blah`. I think it's not much better than 1. We still have to configure the same number of tools, it's just that there's one item in the config.

Personally I think it's easiest to go with (2). It's a little less convenient when refactoring, but I think overall a smaller maintenance burden. But I don't have experience with shipping NPM packages, and I don't know what other larger JS libraries do.

radex:
Closes #20.

Rewrites all internal imports in 🍉 to be relative paths, which is kinda ugly, and might sometimes be annoying, but simplifies tooling a lot:

- apps can understand 🍉 directly as long as they parse ES7
- … which means no need to rewrite source to export it for external apps' Flow (also avoids potential bugs since Babel doesn't always preserve comment position right, breaking `// $FlowFixMe`)
- … if apps (like we) want to transpile 🍉 on their own (to take advantage of better babel settings), it's easy now
- some ESLint rules now work
- no extra Jest config necessary
- no extra Flow config necessary
- no weird Haste package.json hacks for Metro (integration tests)


radex:
@rozPierog CR plz, then to Marcin

radex:
did a quick purple check - seems ok

src/adapters/lokijs/worker/encodeQuery/index.js
```diff
@@ -17,13 +17,13 @@ import {
...
-import identical from 'utils/fp/identical'
-import objOf from 'utils/fp/objOf'
-import zip from 'utils/fp/zip'
-import cond from 'utils/fp/cond'
-import invariant from 'utils/common/invariant'
+import identical from '../../../../utils/fp/identical'
+import objOf from '../../../../utils/fp/objOf'
+import zip from '../../../../utils/fp/zip'
+import cond from '../../../../utils/fp/cond'
+import invariant from '../../../../utils/common/invariant'
```
mobily: uhh… 😑

radex: yep :/

radex:
@mobily done

mobily:
merging, thanks @radex :)

radex:
yay 🎉

PR #47, opened 2018-09-21
Simplify distribution format

### LINKED ISSUE #22: [Discussion] Better dist format
Currently, we build and output Watermelon in two formats: ESM and CJS. If I understand correctly (@mobily knows this), it's because React Native needs CJS, but Webpack can do tree shaking with ESM. In #19, we add another format (same as source but with rewritten imports) to support Flow.

This complicates things because users have to include a Babel plugin to rewrite `@nozbe/watermelon/xxx` imports to `@nozbe/watermelondb/{esm,cjs}/xxx`.

1. Do we really need ESM for tree shaking? Any 🍉 user is almost surely going to be using almost all features, so I can't see how tree shaking could help. And all potentially tree-shakable things like adapters already require a separate import (`import from '@nozbe/watermelondb/adapters'`) and are not exported from the main import
2. Can we just use CJS for all platforms and remove the need for a babel plugin?
3. We could still write the source code (if we fix #20) in dist, so that an advanced user could set up Babel, tree shaking, whatever they need in their app (We're probably going to do this at Nozbe, since we don't need e.g. `class` transpilation for React Native)

radex:
This changes the distribution output format from:

```
dist/
  esm/
    ... all of watermelon in ES modules format
  cjs/
    ... all of watermelon in common JS format
  src/
    ... watermelon source (for Flow and for custom Babel translations in apps)
```

to:

```
dist/
  ... all of watermelon in common JS format
  src/
    ... watermelon source (for Flow and for custom Babel translations in apps)
```

Here's why I think this is better:

1. This simplifies the build setup and build times a lot (no need to transpile twice)
2. CJS is the most universal format, it Just Works™ everywhere, so we don't have to test as much for different environments (webpack, metro, jest)
3. We no longer have to generate path mappings and ship a Babel plugin. This makes Watermelon simpler to set up for users. And we know from experience with RxJS, that custom path remapping is a pain in the ass
3. We can improve Flow support by just adding `.js.flow` files to folders (no Flow setup will be necessary in apps to rewrite paths)
4. The only benefit from ESM is that Webpack can tree-shake it, but I would argue that:
   - there's no real benefit from it today, because almost all features are used and imported almost all of the time (especially for apps other than "hello world")
   - even if you get a tiny benefit from tree-shaking, we still ship a lot of JUNK (Babel helpers, unnecessary transpilations). So if you want to squeeze as much as possible out of Watermelon, you should be transpiring Watermelon **from source** in your app anyway (we should document how to do this).

Closes #22


radex:
⚠️ don't merge — dependent branch

@mobily CR? ;)

mobily:
@radex could you resolve the conflicts, please?

radex:
✅ conflicts resolved

PR #61, opened 2018-09-26
Refactor Collection `.create` `._update` to use `batch`

rozPierog:


src/Collection/index.js
```diff
@@ -113,9 +112,7 @@ export default class Collection<Record: Model> {
   }
 
   async _update(record: Record): Promise<void> {
-    record._hasPendingUpdate = false
-    await this.database.adapter.batch([['update', record]])
-    this._onRecordUpdated(record)
+    await this.database.batch(record)
```
radex: that's silly ;) This method isn't doing anything now. Change it to use database.batch at the source — in `Model.update`

src/Collection/test.js
```diff
@@ -184,6 +184,9 @@ describe('watermelondb/Collection', () => {
     const m1 = new MockTask(collection, { id: 'm1' })
```
radex: this test should be deleted altogether, since Collection doesn't have any real update logic

1. `Database/test` should check if `database.batch(record.prepareUpdate))` does all the things it should (compare this test with the existing test and add the missing checks if necessary)
2. `Model/test` should have a test (if it doesn't already) that when you do `record.update(x)`, both `record.prepareUpdate` was called (with `x` as arg) and `database.batch` (with `record` as arg)

the idea is to have tests cover all paths but without multiple tests testing the same thing in different places

src/Model/test.js
```diff
@@ -127,6 +127,7 @@ describe('watermelondb/Model', () => {
   it('can update a record', async () => {
     const database = makeDatabase()
     database.adapter.batch = jest.fn()
```
radex: why is this line even here? Is it necessary? is `database` a mock? or is it a real `Database` object?

radex:
thanks @rozPierog 🎉

PR #62, opened 2018-09-26
[RFC] Migrations — specification

radex:
Step 1 in implementing migrations. Here's the API I came up with:

```js
// model/migrations.js
export default schemaMigrations({
  minimumVersion: 1,
  currentVersion: 3,
  migrations: [
    {
      from: 2,
      to: 3,
      steps: [
        addColumn({
          table: 'posts',
          column: { name: 'subtitle', type: 'string', isOptional: true },
        }),
        createTable({
          name: 'comments',
          columns: [
            { name: 'post_id', type: 'string', isIndexed: true },
            { name: 'body', type: 'string' },
          ],
        }),
      ],
    },
    {
      from: 1,
      to: 2,
      steps: [
        // ...
      ],
    },
  ],
})

// index.js
import schema from 'model/schema'
import migrations from 'model/migrations'

const adapter = xxxAdapter({
  schema,
  migrations,
})
```

The version numbers (to/from, minimumVersion, currentVersion) are a little repetitive, but the idea is that this should be as explicit as possible so it can be validated if the migrations cover the entire schema version range correctly.

I'll first implement two types of migration steps: createTable and addColumn. This covers 90% of the cases. You don't _need_ to destroy tables or columns (you can just ignore them), or rename them. And it's relatively rare to change indexing/optionality of columns. In the future, we'll add more migration step types if we need them (or, the community might contribute).

I sketched out future API for them though:

```js
// table operations
destroyTable('table_name')
renameTable({ from: 'old_table_name', to: 'new_table_name' })

// column operations
renameColumn({ table: 'table_name', from: 'old_column_name', to: 'new_column_name' })
destroyColumn({ table: 'table_name', column: 'column_name' })

// indexing
addColumnIndex({ table: 'table_name', column: 'column_name' })
removeColumnIndex({ table: 'table_name', column: 'column_name' })

// optionality
makeColumnOptional({ table: 'table_name', column: 'column_name' }) // allows nulls now
makeColumnRequired({ table: 'table_name', column: 'column_name' }) // nulls are changed to null value ('', 0, false)
```

**alternatives considered**

We could scope migration steps per-table, for example:

```js
{
  from: .., to: ..,
  changes: [
    modifyTable('table_name', [
      rename({ to: 'new_table_name' }),
      addColumn({ name: 'x', type: 'string' }) 
    ])
  ]
}
```

… but this complicates the implementation, without an obvious benefit. You don't write migrations every day, so the API doesn't have to be beautiful, it has to be as foolproof as possible.

WDYT?


kokusGr:
@radex For first iteration I think it is really good :) Some feedback:
1. Why reversed chronological order?
2. Maybe we could later a add simple script to generate migrations as this is kinda boilerplate'y and it's easy to make a mistake (lots of invariants :P)?
3. Alternative approach (`modifyTable`) looks good but if it is hard to implement then maybe support adding multiple columns. Something like this:
```
addColumns({
  table: 'posts',
  columns: [
    { name: 'subtitle', type: 'string', isOptional: true },
    { name: 'thumbnail', type: 'string' },
  ]
})
```
I think this would be the most common use case, but maybe I am wrong :)

radex:
> Why reversed chronological order?

Hmm, I just thought it's most convenient to add new migrations to the TOP of the file, and not the bottom (of a potentially biiiig file)

> Maybe we could later a add simple script to generate migrations as this is kinda boilerplate'y

Very boilerplate'y. But automatically generating migrations based on changed schema would be A LOT of work.

> and it's easy to make a mistake (lots of invariants :P)?

The invariant are precisely for you NOT to make a mistake. If you make a mistake, you'll catch it easily during development, and not by getting thousands of emails from angry customers that the app doesn't work because a migration failed ;)

> Alternative approach (modifyTable) looks good but if it is hard to implement then maybe support adding multiple columns. Something like this:

+1, `addColumns` makes sense!

kokusGr:
> The invariant are precisely for you NOT to make a mistake.

Yes, i know :) Just making a point that it takes a lot of `invariants` to make this safe and I can imagine dev seeing quite a number of red screens writing migrations ;) 

> Very boilerplate'y. But automatically generating migrations based on changed schema would be A LOT of work.

Yes I would imagine so. That's way I said later :)

radex:
> Yes, i know :) Just making a point that it takes a lot of invariants to make this safe and I can imagine dev seeing quite a number of red screens writing migrations ;)

They should be using Flow! 😅

rozPierog:
First of all: Migration api looks sleek! I love how easy to read it is.
Alternative approach may be a bit overkill but `addColumns` suggested by @kokusGr makes total sense, but I think it would be great to expand it to at least `destroyColumns` just because I think that it wouldn't be that hard to do and would allow for much more consistent migration code.

radex:
> but I think it would be great to expand it to at least destroyColumns just because I think that it wouldn't be that hard to do and would allow for much more consistent migration code.

it's not trivial. SQLite doesn't support deleting columns, so you have to do a transaction where you repackage data between old and new tables. Likewise, LokiJS is schema-less, so you have to rewrite all records to remove a field. One rarely removes columns, and when you do, you can just ignore their contents, so I think it's safe to ignore it _for now_

radex:
@kokusGr Changed `addColumn` to `addColumns`. CR please

kokusGr:
LGTM :)

febeling:
@radex onequestion:

Do you plan to add a general-purpose `execute` command, update callback or some other mechanism to move and update data itself (as opposed to schema) from existing places to potentially newly introduced tables or columns? 

I'm thinking of a situation where you might change a association (relation?) from one-to-many to many-to-many, which would require a new join-table, to give one example.

febeling:
> I think you'd be able to work around this issue in app code (at launch, check if migrations were just ran, and if so, manually move data around).

I see. Still, it might be simpler to receive this information first-hand after the migration ran, and from the migration mechanism, instead of figuring this out. Maybe even by means of custom tracking code. At least some information, e.g. about which versions were migrated across is still authoritatively around. Maybe this and whatever other meta info might come up can be returned as a promise for such an update step to consume.

Just an idea. You should definitively keep your approach of going with the simplest solution first.

radex:
> I see. Still, it might be simpler to receive this information first-hand after the migration ran, and from the migration mechanism, instead of figuring this out. Maybe even by means of custom tracking code. At least some information, e.g. about which versions were migrated across is still authoritatively around. Maybe this and whatever other meta info might come up can be returned as a promise for such an update step to consume.

👍 Will skip this in the first version, but noted as a (near?) future improvement.

mobily:
The migration API is going to change a bit. We will publish more info soon, but this is a good start. Thank you @radex.

fabianmedina09:
Hi, is pending the _renameColumn_ implementation? I need do this for a migration. It is possible or exist any alternative?

radex:
@fabianmedina09 Hi, PRs are welcome :)

PR #70, opened 2018-10-04
Adapter testing refactoring

radex:
- Refactors how Adapters are tested:
  - cleans up shared tests file
  - removes unsafeClearCachedRecords - this "simulated" DB behavior when app is restarted. but this was kind of a hack. we tested a behavior by just clearing cache  array, not actually reconnecting to the database. This also wouldn't allow for schema setup / migration testing
  - adds `adapter.testClone()` method. It returns a new SQLiteAdapter/LokiAdapter object that connects to the same underlying database. (On Loki, we pass the same LokiMemoryAdapter, for SQLite we pass a magic database path for an in-memory database that you can reconnect to). This much more closely simulates an app restart
- `dbName` option in adapters is now optional (simplifies Watermelon setup for new users)

### DIFF
--- a/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
@@ -14,7 +14,6 @@ class Database(private val name: String?, private val context: Context) {
         SQLiteDatabase.openOrCreateDatabase(
                 context.getDatabasePath("$name.db").path
                         .replace("/databases", ""), null)
-
     }
 
     var userVersion: Int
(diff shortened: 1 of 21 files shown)


radex:
@rozPierog check out the code. Android tests currently fail, because they don't understand how to connect to the shared memory database.

Now, in tests, instead of sending dbName of `'test'` or `null`, we send a magic SQLite path `file:testdbXXXX?mode=memory&cache=shared`. This works out of the box with SQLite, so I think it should also work with Android SQLite API, but kotlin code needs to be updated to open a database with this path, and not empty in-memory database

radex:
@rozPierog huh, nevermind, it actually does work now :| Weird, because it would seem like it's going to go through file system instead of in-memory. Can you just verify that this code doesn't actually create a database on disk?

radex:
@kokusGr CR please ;)

Issue #71, opened 2018-10-04
Require cycles


bsr203:
Hi just noticed this warning on chrome console while debugging.

```
require.js:109 Require cycle: node_modules/@nozbe/watermelondb/Collection/index.js -> node_modules/@nozbe/watermelondb/Query/index.js -> node_modules/@nozbe/watermelondb/observation/observeQuery.js -> node_modules/@nozbe/watermelondb/observation/simpleObserver/index.js -> node_modules/@nozbe/watermelondb/Collection/index.js

Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.

```

Is it possible to avoid this?

thanks.

radex:
yes, `simpleObserver` imports a constant from Collection it shouldn't need to. That constant should be moved to Collection/common.js. Feel free to send a pull request

bsr203:
@radex  please see https://github.com/Nozbe/WatermelonDB/pull/72

radex:
fixed

arahmanali:
`
Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
Require cycle: src/navigation/Navigation.js -> src/navigation/registerScreens.js -> src/containers/Authentication/index.js -> src/components/Authentication/index.js -> src/navigation/Navigation.js

`

PR #72, opened 2018-10-04
avoid cyclic dependency and chrome warning: solves #71

bsr203:
hi. I made the changes you suggested. didn't test locally though :-) getting bit long to get started :-)

### DIFF
--- a/src/Collection/common.js
+++ b/src/Collection/common.js
@@ -1,0 +1,6 @@
+// eslint-disable-next-line import/prefer-default-export
+export const CollectionChangeTypes = {
+  created: 'created',
+  updated: 'updated',
+  destroyed: 'destroyed',
+}
--- a/src/Collection/common.js
+++ b/src/Collection/common.js
@@ -1,4 +1,5 @@
 // @flow
+
 // eslint-disable-next-line import/prefer-default-export
 export const CollectionChangeTypes = {
   created: 'created',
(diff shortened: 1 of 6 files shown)


src/Collection/common.js
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line import/prefer-default-export
```
radex: missing `// @flow` at the top of the file (that should be an ESLint rule if you can find that really quick)

bsr203: I added 
```
 flowtype/require-valid-file-annotation:
    - 2
    - always
```

it didn't give an error on linting. But, when I save the file, it automatically insert insert flow annotation. I am not sure why it doesn't trigger error though.

bsr203: if you are happy with the rule, I can add it. I quickly checked the `eslint-plugin-flowtype` repo and there are few issues to go through. Sorry, I haven't used eslint for a while and don't want to spend too much time. thanks

bsr203: here is the link for that rule https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-require-valid-file-annotation

radex: Ok, nvm then. I’ll check this myself as I’d have to configure tests to be ignored. But please add @flow to this file

bsr203: ok done.

radex:
thanks @bsr203

PR #77, opened 2018-10-08
Migrations -- tests

radex:
- add missing tests for migrations
- fix bugs discovered:
  - fix migrations failing on Loki
  - fix migrations being re-applied on next relaunch on Loki
  - fix default value not being set for new columns on SQLite
- simplify common tests


radex:
@kokusGr CR please. Note that CI fails because this doesn't include Android native changes (Leon's branch) yet

src/adapters/__tests__/commonTests.js
```diff
@@ -451,15 +452,162 @@ export default () => [
...
+
+      expect(await adapter.count(checkTaskColumn('test_string', ''))).toBe(2)
+      expect(await adapter.count(checkTaskColumn('test_string_optional', null))).toBe(2)
+      expect(await adapter.count(checkTaskColumn('test_number', 0))).toBe(2)
+      expect(await adapter.count(checkTaskColumn('test_number_optional', null))).toBe(2)
+      expect(await adapter.count(checkTaskColumn('test_boolean', false))).toBe(2)
+      expect(await adapter.count(checkTaskColumn('test_boolean_optional', null))).toBe(2)
+
+      // check I can use new table and columns
+      adapter.batch([
```
kokusGr: Don't need `await` here?

src/adapters/__tests__/commonTests.js
```diff
@@ -451,15 +452,162 @@ export default () => [
...
+        [
+          'create',
+          new MockTask(
+            {},
+            { id: 't3', test_string: 'hey', test_number: 2, test_boolean_optional: true },
+          ),
+        ],
+      ])
+
+      // check that out-of-range migration was not executed
```
kokusGr: Shouldn't out of range migration throw an error?

kokusGr:
Looks ok, only 2 questions :)

src/adapters/__tests__/commonTests.js
radex: Not sure what you're asking — the point is that database we're testing is already at version 3, so migrations from v2 to v3 must not be executed (and we're testing that the migration was not in fact applied)

src/adapters/__tests__/commonTests.js
radex: Oops! Fixed in #81

src/adapters/__tests__/commonTests.js
kokusGr: OK, my bad :)

PR #78, opened 2018-10-09
Android lazy schema loading

rozPierog:
Same thing as https://github.com/Nozbe/WatermelonDB/pull/67 but for Android


native/android/src/main/java/com/nozbe/watermelondb/DatabaseBridge.kt
```diff
@@ -6,93 +6,181 @@ import com.facebook.react.bridge.ReactContextBaseJavaModule
...
-                )
-        )
-        connections[tag] = driver
-        promise.resolve(true)
+        assert(connections[tag] == null) { "A driver with tag $tag already set up" }
+        val promiseMap = Arguments.createMap()
+
+        try {
+            connections[tag] = Connection.Connected(
+                    driver = DatabaseDriver(
```
kokusGr: I don't think that's needed. `Connection.Connected` takes only 1 arg and writing `driver = DatabaseDriver` is kinda repetitive. This is not Java 😝

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -17,7 +17,33 @@ class DatabaseDriver(context: Context, private val configuration: Configuration)
...
+    constructor(context: Context, dbName: String, schema: Schema) : this(context, dbName) {
+        setUpDatabase(schema)
+    }
+
+    constructor(context: Context, dbName: String, migrations: MigrationSet) :
+            this(context, dbName) {
+        migrate(migrations)
+    }
+
+    constructor(context: Context, dbName: String) {
```
kokusGr: I think that this should be a primary constructor and I am wondering if maybe factory pattern wouldn't be more readable here. What I mean by that is create a companion object that will have methods like `createWithMigrations`, `createWithSchema` and `createIfCompatible` or something like that instead of secondary constructors. WDYT?

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -63,26 +59,32 @@ class Database(private val name: String?, private val context: Context) {
...
+                do {
                     allTables.add(it.getString(index))
-                }
+                } while (it.moveToNext())
             }
         }
         return allTables
     }
 
+    fun inTransaction(function: () -> Unit) = db.transaction { function() }
```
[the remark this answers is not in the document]
rozPierog: I had to do it to be able to make transaction outside of Database class :( Will think about how to resolve this without this method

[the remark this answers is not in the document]
radex: can't you define `transaction = db.transaction`? You just want to expose this outside the class

native/android/src/main/java/com/nozbe/watermelondb/DatabaseBridge.kt
rozPierog: You are right about it not being Java - in Java I wouldn't be able to do something like this! 
I was sceptical about this at first but in the end I decided to implement it to a) increase readability of code, b) make it similar to the solution presented in Swift to decrease needed work when someone is working on both codebases later on

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
rozPierog: I don't think that factory is a good idea. Kotlin has very nice system of named arguments so when someone need to create database with migrations he can just use 
```kotlin
DatabaseDriver(context, dbName, migrations = MigrationSet())
```
with Schema
```kotlin
DatabaseDriver(context, dbName, schema = Schema())
```
and to check if compatible 
```kotlin
DatabaseDriver(context, dbName, schemaVersion = SchemaVersion())
```

If you are using named parameters when creating DatabaseDriver there is no confusion about which kind will be created

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -63,26 +59,32 @@ class Database(private val name: String?, private val context: Context) {
...
+                execute("pragma user_version=0")
+                execute("pragma writable_schema=0")
+            }
 
     private fun getAllTables(): ArrayList<String> {
         val allTables: ArrayList<String> = arrayListOf()
         rawQuery(Queries.select_tables).use {
             it.moveToFirst()
             val index = it.getColumnIndex("name")
             if (index > -1) {
```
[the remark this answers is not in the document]
rozPierog: I'm getting column index to also get string that I need from Cursor so consequence of replacing that `if` with `moveToFirst` and getting rid of `val index` is making `allTables.add(it.getString(index))` less readable.

radex:
@rozPierog CI is failing — can you fix? Also, can you rebase this branch on top of adapter-testing ?

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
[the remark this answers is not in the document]
kokusGr: Oops, sorry missed that 😞

native/android/src/main/java/com/nozbe/watermelondb/DatabaseBridge.kt
kokusGr: 👌

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -59,26 +55,40 @@ class Database(private val name: String?, private val context: Context) {
                 return it.getInt(it.getColumnIndex("count"))
             }
 
-    fun unsafeResetDatabase(): Boolean = context.deleteDatabase(name)
-
-    fun inTransaction(function: () -> Unit) = db.transaction { function() }
+//    fun unsafeResetDatabase() = context.deleteDatabase("$name.db")
```
radex: hm, what happened here? Why was this changed to manually deleting contents of the database?

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -157,13 +170,57 @@ class DatabaseDriver(context: Context, private val configuration: Configuration)
...
     }
 
-    init {
-        setUp()
+    private fun setUpDatabase(schema: Schema) {
+        log?.info("Setting up database with version ${schema.version}")
+        try {
+            unsafeResetDatabase(schema)
+        } catch (e: Exception) {
+            e.printStackTrace()
```
radex: doesn't it just add things to the console? In Swift, this is a fatal error (ie app crashes), because there's really nothing you can do to fix the situation...

src/adapters/sqlite/index.js
```diff
@@ -108,42 +108,37 @@ export default class SQLiteAdapter implements DatabaseAdapter {
   }
```
radex: update NativeBridgeType above

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
rozPierog: There are some problems with `context.deleteDatabase(name)` (mainly - it does not deletes database) until I figure it out I changed it in favour of manual deletions

radex: OK. Can you make a task for this? I think actually deleting the database should be more reliable and safer (the deleted data, potentially sensitive, is not garbage collected immediately)

rozPierog: Sure thing!

radex:
Waiting on previous branches to be merged

PR #80, opened 2018-10-09
[Schema] bool → boolean

radex:
Just realized that schema type `bool` should be named `boolean`, since we follow JavaScript conventions, and that's how JS type is called.

I changed this and added only temporary backwards compatibility (for schema creation, not for migrations).

I think we should merge this before we ship migrations as to cause least amount of pain to users


radex:
⚠️ don't merge

@rozPierog please CR

Note: CI fails because it fails on previous branch

radex:
@kokusGr Good catch — fixed

kokusGr:
👌

radex:
@kokusGr ?

PR #81, opened 2018-10-10
Migrations API refactoring

radex:
- Refactored migration spec API based on feedback from @mobily:
  - `minimumVersion`, `currentVersion` parameters are deleted from `schemaMigrations()`
  - `{ from: 3, to: 4, steps: [] }` is simplified to `{ toVersion: 4, steps: [] }`. Skipping multiple versions is not allowed
  - @mobily suggested `version: x`, but I decided that `toVersion: x` is more obvious, since this isn't version 4 _of the migration_, but a migration of the schema from version 3 _to version 4_.
- Updated tests & refactored code based on those changes


radex:
@kokusGr CR please ;)

src/Schema/migrations/test.js
```diff
@@ -78,94 +97,51 @@ describe('schemaMigrations()', () => {
...
-        currentVersion: 3,
-        migrations: [{ from: 1, to: 2, steps: [] }],
+        migrations: [
+          { toVersion: 5, steps: [] },
+          { toVersion: 4, steps: [] },
+          { toVersion: 2, steps: [] },
+        ],
       }),
-    ).toThrow(/cover schema versions/)
+    ).toThrow(/in reverse chronological order/)
```
kokusGr: Shouldn't this throw something like `gap` or `missing step`?

src/Schema/migrations/index.js
```diff
@@ -83,51 +81,54 @@ export type SchemaMigrations = $Exact<{
...
+      }
+      maxCoveredVersion = toVersion
     })
-    invariant(
-      maxCoveredRange === currentVersion,
-      `Invalid migrations! schemaMigrations() says the current version is ${currentVersion}, but migrations listed only cover schema versions range from ${minimumVersion} to ${maxCoveredRange}. Remember that migrations must be listed in reverse chronological order`,
-    )
   }
+
+  const sortedMigrations = sortMigrations(migrations)
```
kokusGr: Why migrations have to be in a reverse chronological order if we are sorting it anyway? We can first sort it and then check for completeness.

kokusGr:
@radex Looks good, only few questions :)

src/Schema/migrations/index.js
```diff
@@ -83,51 +81,54 @@ export type SchemaMigrations = $Exact<{
...
+  const oldestMigration = head(sortedMigrations)
+  const newestMigration = last(sortedMigrations)
+  const minVersion = oldestMigration ? oldestMigration.toVersion - 1 : 1
+  const maxVersion = newestMigration ? newestMigration.toVersion : 1
```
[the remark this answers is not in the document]
radex: Oh I thought we don't support `?.` in Watermelon but I see it's already used in Relation

src/Schema/migrations/index.js
radex: yeah, artifact of previous implementation. If you think this is going to be annoying, I can change it.

kokusGr: I don't think so. Most people will do it anyway :) But maybe we don't have to force reverse order? Some users might find chronological order better.

radex: OK, fixed

radex:
@kokusGr changed as to not force a particular order

src/Schema/migrations/index.js
```diff
@@ -83,51 +78,56 @@ export type SchemaMigrations = $Exact<{
...
+      const { toVersion } = migration
+      if (maxCoveredVersion) {
+        invariant(
+          toVersion === maxCoveredVersion + 1,
+          `Invalid migrations! Migrations listed cover range from version ${minVersion} to ${maxCoveredVersion}, but migration ${JSON.stringify(
+            migration,
+          )} is to version ${toVersion}. Migrations must be listed without gaps, or duplicates.`,
+        )
+      }
+      maxCoveredVersion = toVersion
```
mobily: I guess we don't need the `maxCoveredVersion` variable, use the `reduce` function instead of `forEach`, please.

radex: Oh, Marcin, you… never miss an opportunity to inject a `reduce` 😂

radex:
@mobily 🏓

mobily:
thanks @radex! 🎉

PR #82, opened 2018-10-10
Migrations cleanup

radex:
- If migrations fail, we fall back to resetting the database. I figured it's better if DB is reset, some unsynced data may be lost, and you have to log in to a server to sync again, than to be left with an app that does not and will not work
- added missing tests:  test that DB is reset if no migrations are available, if migrations fail, or if DB is newer than what app supports
- cleaned up Swift and Kotlin code
- schema setup and migrations are now done in a transaction in SQLite
- clean up Adapter and logging (move as much logging as possible to JS)


radex:
@rozPierog CR please (especially Kotlin code)

src/adapters/lokijs/worker/executor.js
```diff
@@ -257,7 +257,12 @@ export default class LokiExecutor {
 
       if (migrationSteps) {
         logger.log(`[DB][Worker] Migrating from version ${dbVersion} to ${this.schema.version}...`)
-        await this._migrate(migrationSteps)
+        try {
+          await this._migrate(migrationSteps)
+        } catch (error) {
+          logger.error('[DB][Worker] Migration failed. Resetting database instead', error)
```
rozPierog: Woah, woah, woah, that's a little bit radical don't you think? What if someone is using Watermelon as an offline database without syncing with server and migration somehow errored? User is going to lose all of his data without a way to rollback. Frankly I don't really know if this is standard practice  in database migrations or not.

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -31,7 +31,11 @@ class DatabaseDriver(context: Context, dbName: String) {
     }
 
     constructor(context: Context, dbName: String, schema: Schema) : this(context, dbName) {
-        setUpDatabase(schema)
+        try {
+            unsafeResetDatabase(schema)
+        } catch (e: Exception) {
+            e.printStackTrace()
```
rozPierog: I think that leaving error handling to the DatabaseBridge is better idea, then you can propagate error to JS and handle it there. Similar to how you've done `setUpWithMigrations`

src/adapters/lokijs/worker/executor.js
kokusGr: I think @radex is right and this should be default behavior. Maybe we could add a flag if someone wants to use Watermelon as offline-only.

The only downside is that we don't provide sync engine, so Watermelon is not going to be usable (reliably) on its own :/

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
radex: @rozPierog let's leave it for later, ok? there are more unlikely failures (like failure to open database) that currently will just fail/crash/whatever, and don't propagate to JS… so it's a different feature than this

radex:
@rozPierog did you get it to work, or still broken?

rozPierog:
@radex Working on it. Progress is slow. As you can see in Emulator Logcat in travis there is still some problems with migrations

radex:
nice! @rozPierog please don't assign two people to PR at the same time, unless you're seeking feedback

rozPierog:
@radex yup! At least I think so. Tests work so it's a good sign right? 
Sorry about 2 assignees, got over excited about passing test :U

kokusGr:
LGTM :) Only one small thing @rozPierog

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -31,7 +31,11 @@ class DatabaseDriver(context: Context, dbName: String) {
     }
 
     constructor(context: Context, dbName: String, schema: Schema) : this(context, dbName) {
-        setUpDatabase(schema)
+        try {
+            unsafeResetDatabase(schema)
+        } catch (e: Exception) {
+            throw e
```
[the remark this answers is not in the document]
radex: so @rozPierog I should just remove the try..catch since it doesn't do anything, yeah?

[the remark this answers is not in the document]
rozPierog: @radex yup

radex:
fixed

mobily:
LGTM, thank you! 🎉

PR #86, opened 2018-10-11
Paths cleanup

radex:
minor tweaks:

- ~~Schema folder is renamed to lowercase schema. There's no class/type/whatever called `Schema`, so it feels weird, and doing `import { schemaMigrations } from '@nozbe/watermelondb/Schema/migrations` seems weirdly inconsistent. And making a mistake with capitalization will fail only on some systems, so let's be predictable ;)~~
- moved CollectionMap inside Database folder, since it's only an implementation detail of the latter


radex:
@kokusGr 🏓

radex:
Only very minor changes are left

radex:
@kokusGr 🏓

kokusGr:
@rozPierog CI fails again :( From what I can tell emulator API is not compatible?

kokusGr:
@rozPierog CI fails again :( From what I can tell emulator API is not compatible?

radex:
@kokusGr no, those are random failures… can't figure out why they happen.

rozPierog:
@kokusGr This has nothing to do with emulator system versions (as far as I know) Travis is just very shitty when it comes to reliability in Android testing. I ran tests, both as a branch, and merged to master, on my pc and everything is okay.

PR #100, opened 2018-10-19
$ReadOnly

radex:
Adds `$ReadOnly` to a lot of new places. This should fix some of the strange variance issues.

Also, fixes `Q.oneOf()` type from `(string | boolean | number)[]` to `string[] | boolean[] | number[]`


radex:
@kokusGr CR please

src/QueryDescription/index.js
```diff
@@ -23,7 +23,10 @@ import invariant from '../utils/common/invariant'
 
 import { type TableName, type ColumnName, columnName } from '../Schema'
 
+type $RE<Type> = $ReadOnly<$Exact<Type>>
```
kokusGr: We should make it global type and use it everywhere to avoid confusion :)

radex: Agreed, though in a library we can't really make it "global", we'd need to import it anyway from a file. I could start a `src/types.js` file. A lot of types could be moved there too...

kokusGr: 👌 Are you going to do it in this PR?

radex: sure

radex:
@kokusGr 🏓

kokusGr:
LGTM :)

PR #101, opened 2018-10-19
[Travis] restart ADB and retry test on fail

rozPierog:
My mission is to create the most robust Android testing CI script ever. So I've made simple script to just retry test if it fails for the first time. It Should lower rate of false-negatives in CI created by some problem with ADB.  **_IT'S NOT GONNA BE FOOLPROOF_** but still better than nothing - right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ matrix:
         - yarn
       script:
         - TERM=dumb yarn ktlint
-        - android-wait-for-emulator
+        - TERM=dumb android-wait-for-emulator
         - adb kill-server
         - adb start-server
         - adb shell input keyevent 82 &
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,11 +65,9 @@ matrix:
         - yarn
       script:
         - TERM=dumb yarn ktlint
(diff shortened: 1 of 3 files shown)


radex:
So if I understand correctly: If tests fail for any reason, it will retry ONCE, correct? It doesn't have any magic to figure out reason for failure, so a legitimate tests failure will cause the test to (unnecessarily) run twice, yeah? (Not a problem, just not sure if I understand bash correctly)

radex:
you have a syntax error :P

rozPierog:
>if tests fail for any reason, it will retry ONCE, correct? It doesn't have any magic to figure out reason for failure, so a legitimate tests failure will cause the test to (unnecessarily) run twice, yeah? 

Well, yeah. But you've just inspired me to become a wizard and do some magic to only restart if it failed due to `no connected devices` error

radex:
> Well, yeah. But you've just inspired me to become a wizard and do some magic to only restart if it failed due to no connected devices error

i'd avoid wizardry unless you're determined, since there might be other sources of flakiness fixed by this 😂

radex:
thanks!

Issue #60, opened 2018-09-25
Add automatic test coverage measurements


radex:
can Jest output this? do we need a separate tool for it?

would be great if measurements (and diff in test coverage) were output in pull requests. We could hook it up to Danger/Peril maybe.

PR #108, opened 2018-10-24
Danger integration preparation

kkoscielniak:
As mentioned in #60 you wanted to have tests coverage output in PRs. 

I have added the coverage to your `jest.config.js`, it's testing `src/` files only right now (easily changeable) and provides reports in HTML and JSON (the latter can be used for DangerJS integration, it's available in `coverage/coverage-final.js` after running tests). 

As I am not familiar with how did you set up DangerJS and I am not familiar with the software itself I am not touching this issue, but from what I've seen it should be possible to add Jest integration with [this](https://www.npmjs.com/package/danger-plugin-jest) package. 

Btw. In the meantime, I've checked your README with Grammarly and there were 2 typos, which I've fixed :) 

Hope it helps!


radex:
Not bad! This is cool! The problem I can see is Danger messages are done by a Peril server, they're not part of the Travis CI run. So there's no way to easily output the results of test coverage.

The only thing I can think of is to commit the coverage files to the repo… but it's messy + I can't think of a way to guarantee that someone ran tests before committing to update the coverage

radex:
… or I guess we could use https://coveralls.io/features integrated with Travis CI?

I don't have experience with hooking up test coverage. Could use feedback from someone who does

kkoscielniak:
Well, that’s a nice idea to use coveralls. I will dig into it.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 ### Deprecations
 
 - [Schema] Column type 'bool' is deprecated — change to 'boolean'
+- Updated docs
+- Added fundaments for integration of Danger with Jest
```
radex: hm, this is not under the right header

radex:
Hey @kkoscielniak, did you have a chance to take a look at coveralls? I can merge this as-is if you don't want to spend time on this now, but please update CHANGELOG because it got out of sync ;)

kkoscielniak:
@radex I will update the CHANGELOG to enable you to merge. I think tomorrow, would be the time. 

Coveralls will be a separate PR, is that okay?

radex:
Thanks @kkoscielniak !

Issue #112, opened 2018-10-25
Support like and notLike operator


ieschalier:
What do you think about add **like** and **notLike** operator to query ?

It could be a good optimization for applications with search feature in large amount of data.

## Proposal

```js
commentCollection.query(
  Q.where('label', Q.like('a%')),
)
```

radex:
this should be possible, not super difficult, but not trivial either. If someone wants to implement this, I can give guidance as for where to start

ieschalier:
I'm ok to implement it. Can you give me some advice ?

radex:
Sure, you need these parts:

- QueryDescription/index.js is where operators are defined
- adapters/sqlite/encodeQuery is where the query is converted to SQL (react native)
- adapters/lokijs/encodeQuery is the same but for LokiJS (web)
- observation/encodeMatcher is where the query is encoded into a **matcher function**, which can tell whether a `Model` object matches a query or not. This is necessary for observing queries. More details here: https://github.com/Nozbe/WatermelonDB/issues/15#issuecomment-419713849
- Tests! The first three items have low-level unit tests to see if the right output is generated from the query language
- … and then there's the integration test in `__tests__/databaseTests` that checks every single operator on every supported database to see if it actually works

ieschalier:
@radex My first PR to WatermelonDB has been sent 🙂. I'm waiting for your feedback

sonikro:
Hello, great work guys.

While this is not approved yet, is there any alternative way of doing a LIKE query ? I really need this feature to implement a filter by Customer Name in my App.

Thanks.

radex:
@sonikro if it needs to be a fuzzy search… no good solution comes to mind, sorry. I suggest you temporarily use the Watermelon fork @ieschalier has. His PR is almost ready, should work fine — but might take a week or two until it's realeased

sonikro:
Thanks @radex , that's exactly what Im doing right now.

radex:
merged and shipped. cc @sonikro

PR #119, opened 2018-10-26
LIKE query 

ieschalier:
Make choice to use SQL LIKE syntax.
> For lokijs and javascript observer the **LIKE** query is converted to **Regexp**

```js
Q.where('name', Q.like('%value%'))
```

Issue: #112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ All notable changes to this project will be documented in this file.
 
 - Updated Flow to 0.83
 
+### Added
+
+- Query like
+
+
 ### Refactoring
 
 - [WIP] Migrations
--- a/src/QueryDescription/index.js
+++ b/src/QueryDescription/index.js
@@ -38,6 +38,7 @@ export type Operator =
   | 'oneOf'
   | 'notIn'
   | 'between'
+  | 'like'
 
 export type ColumnDescription = $Exact<{ column: ColumnName }>
 export type ComparisonRight =
@@ -160,6 +161,10 @@ export function notIn(values: NonNullValue[]): Comparison {
 // Number is between two numbers (greater than or equal left, and less than or equal right)
 export function between(left: number, right: number): Comparison {
   return { operator: 'between', right: { values: [left, right] } }
+}
+
+export function like(value: NonNullValue[]): Comparison {
+  return { operator: 'like', right: { value } }
 }
 
 export function column(name: ColumnName): ColumnDescription {
--- a/src/QueryDescription/index.js
+++ b/src/QueryDescription/index.js
@@ -163,7 +163,7 @@ export function between(left: number, right: number): Comparison {
   return { operator: 'between', right: { values: [left, right] } }
 }
 
-export function like(value: NonNullValue[]): Comparison {
+export function like(value: Value): Comparison {
   return { operator: 'like', right: { value } }
 }
 
--- a/src/__tests__/databaseTests.js
+++ b/src/__tests__/databaseTests.js
@@ -301,6 +301,18 @@ export const matchTests = [
     matching: [{ id: 'm1' }, { id: 'm2', num1: null }, { id: 'm3', num1: 10 }],
     nonMatching: [{ id: 'n1', num1: 0 }, { id: 'n2', num1: 2 }],
   },
+  {
+    name: 'match like',
+    query: [Q.where('value', Q.like('%ipsum%'))],
+    matching: [
+      { id: 'm1', value: 'Lorem ipsum dolor sit amet,' },
+      { id: 'm2', value: 'Lorem Ipsum dolor sit amet,' },
+    ],
+    nonMatching: [
+      { id: 'n1', value: 'consectetur adipiscing elit.' },
+      { id: 'n2', value: null },
+    ],
+  },
 ]
 
 export const joinTests = [
--- a/src/__tests__/databaseTests.js
+++ b/src/__tests__/databaseTests.js
@@ -302,7 +302,7 @@ export const matchTests = [
     nonMatching: [{ id: 'n1', num1: 0 }, { id: 'n2', num1: 2 }],
   },
   {
-    name: 'match like',
+    name: 'match like (string)',
     query: [Q.where('value', Q.like('%ipsum%'))],
     matching: [
       { id: 'm1', value: 'Lorem ipsum dolor sit amet,' },
@@ -312,6 +312,17 @@ export const matchTests = [
       { id: 'n1', value: 'consectetur adipiscing elit.' },
       { id: 'n2', value: null },
     ],
+  },
+  {
+    name: 'match like (null)',
+    query: [Q.where('value', Q.like(null))],
+    matching: [
+      { id: 'm1', value: 'Lorem ipsum dolor sit amet,' },
+      { id: 'm2', value: 'Lorem Ipsum dolor sit amet,' },
+      { id: 'm3', value: 'consectetur adipiscing elit.' },
+      { id: 'm4', value: null },
+    ],
+    nonMatching: [],
   },
 ]
 
--- a/src/adapters/lokijs/worker/encodeQuery/index.js
+++ b/src/adapters/lokijs/worker/encodeQuery/index.js
@@ -87,9 +87,17 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => valu
   $and: [operator(value), weakNotEqual(null)],
 })
 
-const like = value => ({
-  $regex: new RegExp(value.replace(/%/g, '.*').replace('_', '.'), 'i'),
-})
+const like: OperatorFunction = value => {
+  if (typeof value === 'string') {
+    return {
+      $regex: new RegExp(value.replace(/%/g, '.*').replace('_', '.'), 'i'),
+    }
+  }
+
+  return {
+    $regex: new RegExp('.*', 'i'),
+  }
+}
 
 const operators: { [Operator]: OperatorFunction } = {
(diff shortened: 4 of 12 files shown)


radex:
@kokusGr mind taking a look? (in spare time, of course0

radex:
@ieschalier nice work! Thank you for contributing 🎉

mobily:
this is truly awesome! thank you @ieschalier 💯

src/__tests__/databaseTests.js
```diff
@@ -301,6 +301,29 @@ export const matchTests = [
     matching: [{ id: 'm1' }, { id: 'm2', num1: null }, { id: 'm3', num1: 10 }],
     nonMatching: [{ id: 'n1', num1: 0 }, { id: 'n2', num1: 2 }],
   },
+  {
+    name: 'match like (string)',
+    query: [Q.where('value', Q.like('%ipsum%'))],
```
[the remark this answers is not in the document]
ieschalier: @rozPierog Tests has been fixed 🙂.

src/__tests__/databaseTests.js
```diff
@@ -301,6 +301,29 @@ export const matchTests = [
...
+    name: 'match like (null)',
+    query: [Q.where('text1', Q.like(null))],
+    matching: [],
+    nonMatching: [
+      { id: 'n1', text1: 'Lorem ipsum dolor sit amet,' },
+      { id: 'n2', text1: 'Lorem Ipsum dolor sit amet,' },
+      { id: 'n3', text1: 'consectetur adipiscing elit.' },
+      { id: 'n4', text1: null },
+    ],
+  },
```
kokusGr: I think we need more test like `%value`, `value%`, `value%value`, `_value`, `value_`, `_value%` to be sure that everything works correctly :)

src/adapters/lokijs/worker/encodeQuery/index.js
```diff
@@ -86,6 +87,18 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => valu
   $and: [operator(value), weakNotEqual(null)],
 })
 
+const like: OperatorFunction = value => {
+  if (typeof value === 'string') {
+    return {
+      $regex: new RegExp(value.replace(/%/g, '.*').replace('_', '.'), 'i'),
```
kokusGr: Shouldn't second replace also use regex? `/_/g, '.'` Without you would only replace first occurrence of `_` I think.

src/adapters/lokijs/worker/encodeQuery/index.js
```diff
@@ -86,6 +87,18 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => valu
...
+    }
+  }
+
+  return {
+    $regex: new RegExp('!.*', 'i'),
```
kokusGr: Hmm I am not sure what is the purpose of this? If `value ` is not a string we want to match every word starting from `!`? Maybe I don't understand something :)

kokusGr:
@ieschalier Looks good, only have a few questions + more tests :)

src/adapters/lokijs/worker/encodeQuery/index.js
ieschalier: Good catch! I want to match nothing, but it's not the good way 😅.

My suggestion:
```suggestion
    $regex: new RegExp('(?!.*)', 'i'),
```

ieschalier:
@kokusGr some tests has been added, it highlight bad behavior 😉

src/adapters/lokijs/worker/encodeQuery/index.js
```diff
@@ -86,6 +87,19 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => valu
...
+const like: OperatorFunction = value => {
+  if (typeof value === 'string') {
+    const regexp = `^${value}$`.replace(/%/g, '.*').replace(/_/g, '.')
+    return {
+      $regex: new RegExp(regexp, 'i'),
+    }
+  }
+
+  return {
+    $regex: new RegExp('(?!.*)', 'i'),
```
kokusGr: Good idea :) I think `(?!)` should be enough and as this regex does not have dynamic value it is better to use literal. `/(?!)/i`

kokusGr:
@ieschalier Great, thanks :) LGTM 🎉 There is some conflict with base branch and after that you can assign to radex for final CR/merge :)

ieschalier:
Great day ! 🎉

Conflicts resolved, I don't have right to update assignees. But @radex you can make your final CR ;-)

src/adapters/lokijs/worker/encodeQuery/index.js
```diff
@@ -86,6 +87,19 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => valu
...
+  if (typeof value === 'string') {
+    const regexp = `^${value}$`.replace(/%/g, '.*').replace(/_/g, '.')
+    return {
+      $regex: new RegExp(regexp, 'i'),
+    }
+  }
+
+  return {
+    $regex: new RegExp('(?!)', 'i'),
+  }
```
radex: hm, does this even make sense? I thought only string values are permitted as LIKE operand

src/observation/encodeMatcher/operators.js
```diff
@@ -24,6 +24,18 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => (lef
...
+const handleLikeValue = (v, defaultV) => typeof v === 'string' ? v : defaultV
+
+export const like: OperatorFunction = (left, right) => {
+  const leftV = handleLikeValue(left, '')
+  const rightV = handleLikeValue(right, '(?!)')
+  const regexp = `^${rightV}$`
+    .replace(/%/g, '.*')
+    .replace(/_/g, '.')
+
+  return new RegExp(regexp, 'i').test(leftV)
```
radex: half of this function is the same as in LokiJS worker — I think a util function (in utils/fp) that turns a LIKE query string into a RegExp would be appropriate

ieschalier:
@radex 🤔 good catch. I don't have time last week to update my PR. But I will try to find a better way in next few days.

And also update doc 😉

ieschalier:
@radex I follow your idea of `Q.sanitizeLikeString` it allow only **[a-zA-Z0-9]**.

Actually I drop unwanted character so `hello.*world` become `helloworld`.

> Other solution is to replace all unwanted charactere by `_` : `hello.*world` become `hello__world`.

src/utils/regexp/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as likeToRegexp } from './likeToRegexp'
```
radex: no, make a file in `src/utils/fp/likeToRegexp/index.js`

src/QueryDescription/index.js
```diff
@@ -163,10 +163,14 @@ export function between(left: number, right: number): Comparison {
   return { operator: 'between', right: { values: [left, right] } }
 }
 
-export function like(value: Value): Comparison {
+export function like(value: string): Comparison {
   return { operator: 'like', right: { value } }
 }
 
+export function sanitizeLikeString(value: string): string {
+  return value.replace(/[^a-zA-Z0-9]/g, '')
```
radex: I think replacing unwanted characters with `_` is a better option, otherwise if someone searches "Foo Bar!" it won't find a thing called precisely "Foo Bar!".

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 
 - Added DatabaseProvider and withDatabase Higher-Order Component to reduce prop drilling
 - Added experimental Actions API. This will be documented in a future release.
+- Query like
```
radex: can you merge master to this and update changelog?

src/utils/regexp/likeToRegexp.js
```diff
@@ -0,0 +1,7 @@
+export default likeQuery => {
```
radex:
```suggestion
// @flow 

export default function likeToRegexp(likeQuery: string): RegExp {
```

src/observation/encodeMatcher/operators.js
```diff
@@ -24,6 +25,15 @@ const noNullComparisons: OperatorFunction => OperatorFunction = operator => (lef
 // Same as `a > b`, but `5 > undefined` is also true
 const weakGt = (left, right) => left > right || (left != null && right == null)
 
+const handleLikeValue = (v, defaultV) => typeof v === 'string' ? v : defaultV
+
+export const like: OperatorFunction = (left, right) => {
+  const leftV = handleLikeValue(left, '')
+  const rightV = handleLikeValue(right, '(?!)')
```
radex: i don't think this is correct. I _think_ right cannot be null, and also the fallback `?!` will be replaced to nothing by likeToRegexp

src/QueryDescription/index.js
```diff
@@ -165,6 +166,14 @@ export function between(left: number, right: number): Comparison {
   return { operator: 'between', right: { values } }
 }
 
+export function like(value: string): Comparison {
+  return { operator: 'like', right: { value } }
+}
+
+export function sanitizeLikeString(value: string): string {
```
radex: we need tests for this. In databaseTests, we should test the behavior of querying with a sanitizedLikeString (where the string is something nasty, containing %, _, and regex-unsafe expressions)

radex:
Thanks @ieschalier ! I'll take a look on Friday. Please note that CI is failing — there's a Flow issue

ieschalier:
🎉 

Thank to all team for your help and your advice 😄

PR #125, opened 2018-11-01
Upgrade React Native in native example #trival

azundo:
React native 0.57.0 is incompatible with using decorators in release
mode. Upgrade to the latest 0.57.x which resolves this issue.


radex:
huh, that's weird, I just updated the dependencies

radex:
Thanks!

PR #131, opened 2018-11-06
Actions - Part 1

radex:
Very early proof-of-concept implementation of Actions. This only implements optional locking/sequencing/queueing using `database.action()`, so that we can verify this API in Purple.

There's nothing to force user to use it, no subAction() support, no decorators (if we'll go that route), etc.

(To be continued)


radex:
@kokusGr CR

kokusGr:
LGTM :)

mobily:
great work, thank you! 🎉

PR #136, opened 2018-11-07
Actions — Part 2

radex:
Adds a database constructor flag `actionsEnabled: boolean`. If `true`, you can only perform write operations (create/update/delete) from inside an action. Plus tests.

Ultimately, this will always be enabled, but this way we can experiment and iterate without forcing changes onto users before we're sure this is the right way to go

plus some tests refactoring


radex:
@kokusGr CR

src/Database/index.js
```diff
@@ -88,4 +93,8 @@ export default class Database {
       collection.unsafeClearCache()
     })
   }
+
+  _ensureInAction(error: string): void {
```
kokusGr: Hmm is this safe? I mean there might be other actions in queue and I could use `Collection.create` outside of an action and this wouldn't throw an error. Am I right? :P Sometimes it's kinda hard to follow the flow in async code :/

src/Model/test.js
```diff
@@ -279,6 +263,96 @@ describe('watermelondb/Model', () => {
...
+    await expectToRejectWithMessage(
+      model.markAsDeleted(),
+      /can only be called from inside of an Action/,
+    )
+
+    // check that no throw inside action
+    await database.action(async () => {
+      await model.update(noop)
+      await model.markAsDeleted()
+      await model.markAsDeleted()
```
kokusGr: I think one of them should be `destroyPermanently` :)

radex: nice catch!

src/Database/index.js
radex: You're right! Asynchronicity is really hard! This solution does not give you 100% guarantee of correctness, but you should be able to catch 99% errors, because you'd need to constantly have a long-running action to not hit the error.

I think the only way to guarantee correctness would be to remove batch/create/update methods, and have those methods only on an object passed to the `database.action()` block

kokusGr: Ok, thanks for explanation :) I guess 99% is good enough for now :)

PR #140, opened 2018-11-08
Subactions

radex:
adds subaction API, as discussed to be able to call an action from an action, skipping the queue:

```
// direct api
database.action(async action => {
  action.subAction(() => anotherAction())
})

// @action api in Model instances
@action async foo() {
 await  this.subAction(() => this.bar())
}
```


radex:
@kokusGr CR please :)

kokusGr:
LGTM :)

PR #144, opened 2018-11-09
Fix type in Migrations.MD

HishamMubarak:
Without migrations, if a user of your app upgrades from one version to another, their local database will be cleared at launch, and they will "use" => "lose" all their data.

Changed use to lose in the above sentence. I am not 100% sure if I am correct, maybe this is a feature and I'm mistaken. If so, please reject this PR.


radex:
whoa, good catch. This is a particularly bad word to screw up! Thanks!

PR #146, opened 2018-11-13
Separate arrayDifference

lekterable:
I hope I didn't go too far with the tests, there is never too much of them, right?
I've also assumed that objects in line 31 and 34 are not equal on purpose.

### DIFF
--- a/src/utils/fp/arrayDifference/index.js
+++ b/src/utils/fp/arrayDifference/index.js
@@ -1,7 +1,7 @@
 // @flow
 import withoutIdentical from '../withoutIdentical'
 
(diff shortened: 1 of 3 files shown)


src/utils/fp/arrayDifference/index.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import withoutIdentical from '../withoutIdentical'
+
+type ArrayDiff<T> = { added: T[], removed: T[] }
```
radex:
```suggestion
export type ArrayDiff<T> = $Exact<{ added: T[], removed: T[] }>
```

radex:
thanks @lekterable !

PR #150, opened 2018-11-14
Sync primitives — Part 1

radex:
Some basic preparation/clean up for sync primitives:

- move existing helpers around + add tests
- add basic sync helper (unused yet) + tests

### DIFF
--- a/src/sync/helpers.js
+++ b/src/sync/helpers.js
@@ -1,0 +1,19 @@
+// @flow
+
+import type { ColumnName } from '..'
+import type { RawRecord } from '../RawRecord'
+
+export function addToRawSet(rawSet: ?string, value: string): string {
+  const array = rawSet ? rawSet.split(',') : []
+  const set = new Set(array)
+  set.add(value)
+  return Array.from(set).join(',')
+}
+
+// Mutates `rawRecord` to mark `columName` as modified for sync purposes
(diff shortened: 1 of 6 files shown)


radex:
@kokusGr CR please :)

src/sync/helpers.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import type { ColumnName } from '..'
+import type { RawRecord } from '../RawRecord'
+
+export function addToRawSet(rawSet: ?string, value: string): string {
+  const array = rawSet ? rawSet.split(',') : []
+  const set = new Set(array)
+  set.add(value)
+  return Array.from(set).join(',')
```
kokusGr: Hmm this seems like a lot of work. Maybe we should think about other soultion. But I guess first we need to test it's performance ;) Anyway it's not the most important thing to do right now.

kokusGr:
LGTM :) But there is a conflict

src/sync/helpers.js
radex: this code is unchanged, just moved to a different place ;) I'll be tweaking performance once the whole sync infra is done.

kokusGr: Yes, i know just wanted to point that out ;)

mobily:
thanks @radex! :)

PR #151, opened 2018-11-14
Sync primitives — Part 2

radex:
Adds two building blocks necessary for sync:

- function for fetching local changes (that need to be pushed to the server to sync)
- function for marking local changes as synced (after server acknowledges it has the new data)
- tests

Those are not the best/prettiest/most performant implementations ever — I will continue tweaking it in future PRs. I don't want to spend too much time refactoring this code yet, because I might have to completely redo it as I understand all the intricacies of sync.

### DIFF
--- a/src/Database/test.js
+++ b/src/Database/test.js
@@ -27,7 +27,8 @@ describe('Database', () => {
 
 describe('Batch writes', () => {
   it('can batch records', async () => {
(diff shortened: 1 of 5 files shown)


radex:
@kokusGr CR please :)

kokusGr:
LGTM :)

mobily:
thanks, merging! :)

PR #160, opened 2018-11-21
[WIP] Sync primitives - Part 5

radex:
- Fixes an issue where `create` server changes wouldn't get applied if record is locally mistakenly marked as `deleted`
- Fixes issues relating to automating updated_at/created_at timestamps


radex:
@kokusGr CR please :)

PR #161, opened 2018-11-22
Add docs for observing 2nd level relations to Components.md

jon-dearaujo:
Including new advanced section to docs/Components.md describing how to observe 2nd level relations.

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -194,18 +194,6 @@ const enhanceAuthorContact = withObservables(['author'], ({author}) => ({
 
 const EnhancedPost = enhancePostAndAuthor(enhanceAuthorContact(PostComponent));
 ```
-With this approach, `PostComponent` will have `Post`, `Author` and `Contact` props.
-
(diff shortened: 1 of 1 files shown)


jon-dearaujo:
@radex Please, take a look.

docs/Components.md
```diff
@@ -178,6 +178,50 @@ const EnhancedCommentList = enhance(CommentList)
...
+
+If you are familiar with `rxjs`, another way to achieve the same result is using `switchMap` operator:
+
+```js
+import { switchMap } from 'rxjs/operators'
+
+const enhancePost = withObservables(['post'], ({post}) => ({
+  post: post.observe(),
+  author: post.author.observe(),
+  contact: post.author.observe().pipe(switchMap(author => author.contact.observe()))
```
radex: that's confusing. I believe you meant to say that this is the same as the original withObservables composition, but when reading, it seems like another solution to the problem of optional relations, and it's not (it will error out, because there's no `contact` on `null`).

jon-dearaujo: Sure, it is confusing. My idea was to suggest one alternative for the 2nd level, without the optional relation. Perhaps the order made it confusing, and placing it above the optional relation note would make it more clear. Let me try it.

jon-dearaujo: @radex I reordered some of the text, I think it makes more sense now.

PR #162, opened 2018-11-22
Sync primitives - Part 6

radex:
We're getting close!

- Clean up files: split tests file; move sync/index to sync/impl (GH can't figure it out and shows a big diff - sorry! there's no actual changes in impl.js except getLastSyncedAt/setLastSyncedAt added)
- Add an implementation of top-level `synchronize()` function that combines all sync phases together 🎉 
- Tests for `synchronize()`
- Clean up tests


radex:
@kokusGr CR please :)

Issue #141, opened 2018-11-08
⚡️ Easy issues for first-time contributors! 🍉


radex:
Hey everyone! There’s a lot of you out there interested in using Watermelon! You can help us a lot in developing 🍉 by contributing to it — even in simple ways!

Here’s a list of easy fixes/improvements you can do to help:

- [ ] WIP

lekterable:
Screw sleep, I was really curious about the code so I've did:

> Move arrayDifference from observation/fieldObserver to utils/fp/arrayDifference and add a couple of tests

Opening a PR for it in a minute.

radex:
@lekterable Nice! Thank you for the suggestion — this list is written up definitely thanks to you

cesargdm:
So talking about hooks, and taking advantage of its stable release. For example, ¿the way It should be to use hooks in observables could be?:

```javascript
import useObservable from '@nozbe/use-observable'

function Comment() {
  const { comment } = useObservables(['comment'], ({ comment }) => ({
    comment: comment.observe()
  }))
  
  return (  
    <div>
      <p>{comment.body}</p>
    </div>
  )
}

function Post() {
  const { post, comments } = useObservables(['post'], ({ post }) => ({
    post: post.observe(),
    comments: post.comments.observe()
  }))

  return (
    <article>
      <h1>{post.name}</h1>
      <p>{post.body}</p>
      <h2>Comments</h2>
      {comments.map(comment =>
        <EnhancedComment key={comment.id} comment={comment} />
      )}
    </article>
  )
}

```

radex:
Yes! There's a proof of concept implementation of a useObservable on Nozbe/withObservables repo, but it hasn't been committed... A pull request would be nice

co16353sidak:
Hey Everyone, I would love to work on updating the docs for watermelonDB regarding the `withObservables()` issue, would love it if I got the go ahead and a some tips on how to do this. 🥇

radex:
@co16353sidak Hey, I _think_ the withObservables docs have been updated — but feel free to read through the docs carefully again and submit a PR with tweaks and fixes when you feel like something is not exactly understandable

co16353sidak:
Is the issue regarding removing depreciated babel plugin still open?, would love to try this one out ?
@radex

radex:
@co16353sidak go for it!

saqibameen:
[...]

PR #163, opened 2018-11-23
 Add docs for Many to Many relation

### LINKED ISSUE #73: Many to many Relationships
Thank you for your awesome work on this library!

My question:
Is there support (planed?) for many-to-many relationships?

Example:
I have a cookbook table. A cookbook belongs to a specific user.
I have a recipe table.
A recipe can belong to many cookbooks (a user can „favorite“ the recipe in his personal cookbook).

This could be solved using a associative/pivot table containing foreign keys to the cookbook table and the recipe table.

Kabangi:
Closes #73

### DIFF
--- a/docs/Model.md
+++ b/docs/Model.md
@@ -111,7 +111,7 @@ class Comment extends Model {
 
 **➡️ Learn more:** [Relation API](./Relation.md)
 
-### Children
+### To-Many relation
 
 To point to a list of records that belong to this Model, e.g. all `Comment`s that belong to a `Post`, you can define a simple `Query` using `@children`:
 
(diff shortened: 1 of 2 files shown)


docs/Model.md
```diff
@@ -111,7 +111,7 @@ class Comment extends Model {
 
 **➡️ Learn more:** [Relation API](./Relation.md)
 
-### Children
+### To-Many relation
```
radex:
```suggestion
### Children (To-Many relation)
```

docs/Model.md
```diff
@@ -132,6 +132,66 @@ Pass the _table name_ of the related records as an argument to `@children`. The
 
 **Note:** You must define a `has_many` association in `static associations` for this to work
 
+### Many-To-Many Relation
```
radex: I suggest moving it down to the "Advanced" section

docs/Model.md
```diff
@@ -132,6 +132,66 @@ Pass the _table name_ of the related records as an argument to `@children`. The
...
+
+If for instance, our app `Post`s can be authored by many `User`s and a user can author many `Post`s. We would create such a relation following these steps:- 
+
+1. Create a pivot schema and model that both the `User` model and `Post` model has association to; say `PostUser`
```
radex: PostUser → PostAuthor perhaps? i think that would be more clear

docs/Model.md
```diff
@@ -132,6 +132,66 @@ Pass the _table name_ of the related records as an argument to `@children`. The
...
+import {lazy } from '@nozbe/watermelondb/decorators'
+
+class Post extends Model {
+  static table = 'posts'
+  static associations = {
+    post_users: { type: 'has_many', foreignKey: 'post_id' },
+  }
+
+  @lazy
+  posts = this.collections
```
radex:
```suggestion
  authors = this.collections
```

docs/Model.md
radex: I think that's a little more clear

radex:
> Also, I think we should migrate the relation docs from model docs to Relation docs for better visibility. If this okay I can go ahead and do that.

Ah, you're right. I think the very basics ( @children, relation) should stay as is, but the advanced details (like many-to-many ) should be moved to Relation.md

docs/Model.md
Kabangi: Yes much better

docs/Model.md
Kabangi: > > Also, I think we should migrate the relation docs from model docs to Relation docs for better visibility. If this okay I can go ahead and do that.
> 
> Ah, you're right. I think the very basics ( @children, relation) should stay as is, but the advanced details (like many-to-many ) should be moved to Relation.md

Yes, I think this is much better. I will move the docs there then.

Kabangi:
@radex Updated

radex:
PS. Thanks again @Kabangi for contributing. Would you like to play a little more with Watermelon's code or docs? I have some a list of easy things to tackle: https://github.com/Nozbe/WatermelonDB/issues/141 if you'd like to contribute. (And I can offer to send you some Watermelon stickers ;) )

docs/Model.md
```diff
@@ -132,7 +132,7 @@ Pass the _table name_ of the related records as an argument to `@children`. The
 
 **Note:** You must define a `has_many` association in `static associations` for this to work
 
-**➡️ Learn more:** [Queries](./Query.md)
+**➡️ Learn more:** [Relation API](./Relation.md)
```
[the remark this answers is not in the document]
Kabangi: I will revert, I missed the context here a bit.

Kabangi:
@radex Am happy to do both code and docs. I will be looking to send more PR's next week

PR #167, opened 2018-11-27
Sync primitives — Part 7

radex:
- Forces Actions to be enabled in Watermelon to use Sync (otherwise it's unsafe)
- Makes applyRemoteChanges faster: it now runs more concurrently, and updates/creates are in one big batch
   - sadly, it makes code a lot harder to follow :( … but it's the best I came up with so far
   - refactor ideas appreciated ❤️

### DIFF
--- a/src/sync/impl.js
+++ b/src/sync/impl.js
@@ -11,8 +11,9 @@ import {
   filter,
   find,
   equals,
+  piped,
 } from 'rambdax'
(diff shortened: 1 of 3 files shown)


mobily:
> sadly, it makes code a lot harder to follow :(

it doesn't look as bad as you think 😎 thanks, merging!

PR #169, opened 2018-11-28
Sync: Documentation

radex:
- Small changes (as discussed): lastSyncedAt → lastPulledAt, and pass this to pushChanges() as well
- Add documentation:
  - user's guide / server-side implementation guide + TODO (this will be moved to docs/ markdown file when we're ready)
  - implementation documentation - for future generations, since this is complicated stuff and hard to understand all important nuances

### DIFF
--- a/src/sync/index.js
+++ b/src/sync/index.js
@@ -35,6 +35,133 @@ export type SyncArgs = $Exact<{
   pushChanges: SyncPushArgs => Promise<void>,
 }>
 
+/*
+
+# Watermelon synchronization protocol
+
+To synchronize, you need to pass two functions, `pullChanges` and `pushChanges` that can talk to your
+backend.
+
+Changes are represented as an object with *raw records*. Those only use raw table and column names,
+and raw values (strings/numbers/booleans) — the same as in Schema. Deleted objects are always only
+represented by their IDs.
+
+Example:
+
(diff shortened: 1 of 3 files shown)


radex:
@tkapelak Read through the documentation please and let me know if I forgot about something important or did something contrary to what we decided on.

src/sync/index.js
```diff
@@ -24,34 +24,246 @@ export type SyncDatabaseChangeSet = $Exact<{ [TableName<any>]: SyncTableChangeSe
...
+Changes are represented as an object with *raw records*. Those only use raw table and column names,
+and raw values (strings/numbers/booleans) — the same as in Schema. Deleted objects are always only
+represented by their IDs.
+
+Example:
+
+```
+{
+  projects: {
+    created: [
```
tkapelak: The backend implementation we are woking on doesn't allow to detect if the record has been changed or created since the last sync. It would require to add `created_at` field to all records in collections... But the the procedure for "updated" says "if doesn't exist locally (error), create" so maybe it's not a big deal to send everything (except "deleted") in `updated`?

radex: oh 🤦‍♂️ don't know why I didn't realize this. will make sure this will still work correctly and make appropriate changes

PR #170, opened 2018-11-29
Remove last_modified

radex:
We don't actually need `last_modified` for our new sync implementation. Some people might, but there's no reason why it should be built in if one can just define it in app schema.

### DIFF
--- a/src/RawRecord/test.js
+++ b/src/RawRecord/test.js
@@ -57,7 +57,6 @@ describe('sanitizedRaw()', () => {
       id: 'abcdef',
       _status: 'synced',
       _changed: '',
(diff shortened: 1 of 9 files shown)


radex:
@kokusGr CR please

src/RawRecord/test.js
```diff
@@ -116,12 +113,11 @@ describe('sanitizedRaw()', () => {
       is_all_day: null,
     })
   })
-  it('can create a valid raw from nothin', () => {
+  it(`can create a valid raw from nothin'`, () => {
```
mobily: what happened here? prettier? 😜

PR #172, opened 2018-11-30
Test refactor

radex:


src/adapters/__tests__/commonTests.js
```diff
@@ -55,16 +55,16 @@ export default () => [
             migrations: schemaMigrations({ migrations: [] }),
           }),
       ).not.toThrowError()
```
rozPierog:
```suggestion
      ).not.toThrow()
```
Same on line 45 and 48

radex: you mean whitespace or the `toThrow` → `toThrowError`?

rozPierog: Whitespace change was a mistake, sorry about that. I was talking about changing all `toThrowError` into `toThrow`

rozPierog: > agree with the linter that consistency is more important

Yeah, but something must have gone horribly wrong. Right now there is `toThrow` and `toThrowError` _in the same file_ just few lines of each other. There had to be something wrong with how eslint if as you said it were to replace all names with primary name instead of aliases.

In the _next line_ there is `toThrow` changed from `toThrowError`  That is why I highlighted this line to fix it.

radex: 🤔 ok — will take a look, didn't notice that

radex:
fixed ¯\_(ツ)_/¯

PR #175, opened 2018-12-01
Fix native cache to support non-global IDs

ahmadbaraka:
Changes in this PR address issue #7 to fix the native code to match the JS and support non-global IDs caching

### DIFF
--- a/native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
@@ -167,10 +167,10 @@ class DatabaseDriver(context: Context, dbName: String) {
         log?.info("Mark as cached $id")
         var cache = cachedRecords[table]
         if (cache == null) {
-            cache = mutableListOf<RecordID>()
-            cachedRecords[table] = cache!!
+            cache = mutableListOf()
+            cachedRecords[table] = cache
         }
-        cache!!.add(id)
+        cache.add(id)
     }
 
     private fun isCached(table: TableName, id: RecordID): Boolean = cachedRecords[table]?.contains(id) ?: false
--- a/native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
@@ -9,8 +9,8 @@ import java.util.logging.Logger
 
 class DatabaseDriver(context: Context, dbName: String) {
     sealed class Operation {
-        class Execute(val query: SQL, val args: QueryArgs) : Operation()
-        class Create(val id: RecordID, val query: SQL, val args: QueryArgs) : Operation()
+        class Execute(val table: TableName, val query: SQL, val args: QueryArgs) : Operation()
+        class Create(val table: TableName, val id: RecordID, val query: SQL, val args: QueryArgs) : Operation()
         class MarkAsDeleted(val table: TableName, val id: RecordID) : Operation()
         class DestroyPermanently(val table: TableName, val id: RecordID) : Operation()
         // class SetLocal(val key: String, val value: String) : Operation()
@@ -43,10 +43,10 @@ class DatabaseDriver(context: Context, dbName: String) {
 
     private val log: Logger? = if (BuildConfig.DEBUG) Logger.getLogger("DB_Driver") else null
 
-    private var cachedRecords: ArrayList<String> = arrayListOf()
+    private val cachedRecords: MutableMap<TableName, MutableList<RecordID>> = mutableMapOf()
 
     fun find(table: TableName, id: RecordID): Any? {
-        if (isCached(id)) {
+        if (isCached(table, id)) {
             return id
         }
         database.rawQuery("select * from $table where id == ? limit 1", arrayOf(id)).use {
@@ -54,24 +54,24 @@ class DatabaseDriver(context: Context, dbName: String) {
                 return null
             }
             val resultMap = Arguments.createMap()
-            markAsCached(id)
+            markAsCached(table, id)
             it.moveToFirst()
             resultMap.mapCursor(it)
             return resultMap
         }
     }
 
-    fun cachedQuery(query: SQL): WritableArray {
+    fun cachedQuery(table: TableName, query: SQL): WritableArray {
         log?.info("Cached Query: $query")
         val resultArray = Arguments.createArray()
         database.rawQuery(query).use {
             if (it.count > 0 && it.columnNames.contains("id")) {
                 while (it.moveToNext()) {
                     val id = it.getString(it.getColumnIndex("id"))
-                    if (isCached(id)) {
+                    if (isCached(table, id)) {
                         resultArray.pushString(id)
                     } else {
-                        markAsCached(id)
+                        markAsCached(table, id)
                         resultArray.pushMapFromCursor(it)
                     }
                 }
@@ -129,29 +129,29 @@ class DatabaseDriver(context: Context, dbName: String) {
     }
 
     fun batch(operations: List<Operation>) {
-        val newIds = arrayListOf<RecordID>()
-        val removedIds = arrayListOf<RecordID>()
+        val newIds = arrayListOf<Pair<TableName, RecordID>>()
+        val removedIds = arrayListOf<Pair<TableName, RecordID>>()
         database.transaction {
             operations.forEach {
                 when (it) {
                     is Operation.Execute -> execute(it.query, it.args)
                     is Operation.Create -> {
                         create(it.id, it.query, it.args)
-                        newIds.add(it.id)
+                        newIds.add(Pair(it.table, it.id))
                     }
                     is Operation.MarkAsDeleted -> {
                         database.execute(Queries.setStatusDeleted(it.table), arrayListOf(it.id))
-                        removedIds.add(it.id)
+                        removedIds.add(Pair(it.table, it.id))
                     }
                     is Operation.DestroyPermanently -> {
                         database.execute(Queries.destroyPermanently(it.table), arrayListOf(it.id))
-                        removedIds.add(it.id)
+                        removedIds.add(Pair(it.table, it.id))
                     }
                 }
             }
         }
-        newIds.forEach(this::markAsCached)
-        removedIds.forEach { cachedRecords.remove(it) }
+        newIds.forEach { markAsCached(it.first, it.second) }
+        removedIds.forEach { removeFromCache(it.first, it.second) }
     }
 
     fun unsafeResetDatabase(schema: Schema) {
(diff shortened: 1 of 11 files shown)


radex:
Nice work @ahmadbaraka! 

There are some issues:
- See CI output: https://travis-ci.com/Nozbe/WatermelonDB/jobs/162152637 — Swift doesn't compile and there are Flow errors
- It would be good to have at least a basic test verifying that both adapter.query and adapter.find actually behave correctly. This is the file: https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/__tests__/commonTests.js

ahmadbaraka:
@radex Thanks for quick feedback.
Some changes were done by mistake on the `dev` version in my app instead of being done in the source repo, so they were missed in the commit.

Currently waiting for CI to complete to verify the fixes.

----

> It would be good to have at least a basic test verifying that both adapter.query and adapter.find actually behave correctly. This is the file: https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/__tests__/commonTests.js

Regarding the tests, I see that file has common tests used for the adapters, but the changes here include implementation details of the SQLite adapter. I see the current tests also verify that `query` and `find` would return either `id` if the model is already cached, or `model._raw` if it's not cached.
Could you point out which test cases we can add here?

radex:
> Regarding the tests, I see that file has common tests used for the adapters, but the changes here include implementation details of the SQLite adapter. I see the current tests also verify that query and find would return either id if the model is already cached, or model._raw if it's not cached.
> Could you point out which test cases we can add here?

Right, don't test the implementation details, test the behavior. I'd add a test for both query and find, checking that if I query/find "xxxx" in table A, which is already cached, it will return "xxxx", but if I do the same with table B (where it's not cached), it still returns the whole record.

PS. You only made the change in native implementation, not in LokiJS/web implementation. So you'll have to either also add the LokiJS implementation (should be easy), or, you can just wrap the test in is-native check (in which case, please mark in changelog that this is PARTIAL support for non-global IDs - native only)

ahmadbaraka:
I've pushed additional changes to address your comment:
1. Added the LokiJS implementation for non-global IDs
1. Implemented tests for the new changes for `find` and `query` as suggested

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -129,29 +129,29 @@ class DatabaseDriver(context: Context, dbName: String) {
...
-                        removedIds.add(it.id)
+                        removedIds.add(Pair(it.table, it.id))
                     }
                 }
             }
         }
-        newIds.forEach(this::markAsCached)
-        removedIds.forEach { cachedRecords.remove(it) }
+        newIds.forEach { markAsCached(it.first, it.second) }
+        removedIds.forEach { removeFromCache(it.first, it.second) }
```
rozPierog:
```suggestion
        removedIds.forEach { removeFromCache(table = it.first, id = it.second) }
```

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -163,12 +163,19 @@ class DatabaseDriver(context: Context, dbName: String) {
...
     fun close() = database.close()
 
-    private fun markAsCached(id: RecordID) {
+    private fun markAsCached(table: TableName, id: RecordID) {
         log?.info("Mark as cached $id")
-        cachedRecords.add(id)
+        var cache = cachedRecords[table]
+        if (cache == null) {
+            cache = mutableListOf<RecordID>()
+            cachedRecords[table] = cache!!
```
rozPierog: `mutableListOf<RecordID>()` is creating non-nullable type of object so double bang is unnecessary

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -163,12 +163,19 @@ class DatabaseDriver(context: Context, dbName: String) {
...
+        }
+        cache!!.add(id)
```
rozPierog: Same here

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -163,12 +163,19 @@ class DatabaseDriver(context: Context, dbName: String) {
 
     fun close() = database.close()
 
-    private fun markAsCached(id: RecordID) {
+    private fun markAsCached(table: TableName, id: RecordID) {
         log?.info("Mark as cached $id")
-        cachedRecords.add(id)
+        var cache = cachedRecords[table]
+        if (cache == null) {
+            cache = mutableListOf<RecordID>()
```
rozPierog: Why do you state Type of objects in list if you are leaving it empty?

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -129,29 +129,29 @@ class DatabaseDriver(context: Context, dbName: String) {
...
                         database.execute(Queries.destroyPermanently(it.table), arrayListOf(it.id))
-                        removedIds.add(it.id)
+                        removedIds.add(Pair(it.table, it.id))
                     }
                 }
             }
         }
-        newIds.forEach(this::markAsCached)
-        removedIds.forEach { cachedRecords.remove(it) }
+        newIds.forEach { markAsCached(it.first, it.second) }
```
rozPierog:
```suggestion
        newIds.forEach { markAsCached(table = it.first, id = it.second) }
```

native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt
```diff
@@ -163,12 +163,19 @@ class DatabaseDriver(context: Context, dbName: String) {
...
+            cache = mutableListOf<RecordID>()
+            cachedRecords[table] = cache!!
+        }
+        cache!!.add(id)
     }
 
-    private fun isCached(id: RecordID): Boolean = cachedRecords.contains(id)
+    private fun isCached(table: TableName, id: RecordID): Boolean = cachedRecords[table]?.contains(id) ?: false
+
+    private fun removeFromCache(table: TableName, id: RecordID) { cachedRecords[table]?.remove(id) }
```
rozPierog: Why not:
```suggestion
    private fun removeFromCache(table: TableName, id: RecordID) = cachedRecords[table]?.remove(id)
```

native/ios/WatermelonDB/DatabaseDriver.swift
```diff
@@ -163,14 +163,23 @@ class DatabaseDriver {
...
 
-    private func markAsCached(_ id: RecordId) {
-        cachedRecords.insert(id)
+    private func markAsCached(_ table: Database.TableName, _ id: RecordId) {
+        var cachedSet = cachedRecords[table];
+        if cachedSet == nil {
+            cachedSet = []
+        }
+        cachedSet!.insert(id)
+        cachedRecords[table] = cachedSet
```
radex: Let's do this without force unwrapping. How about:

```
var cachedSet = cachedRecords[table] ?? []
cachedSet.insert(id)
cachedRecords[table] = cachedSet
```

src/adapters/__tests__/commonTests.js
```diff
@@ -112,6 +114,72 @@ export default () => [
...
+      adapter = adapter.testClone()
+
+      const s2 = makeMockTask({ id: 'id2', text1: 'baz', order: 2 })
+      await adapter.batch([['create', s2]])
+
+      // returns cached IDs after create
+      expectSortedEqual(await adapter.query(taskQuery()), [s1._raw, 'id2'])
+
+      // returns raw if not cached for a different table
+      expectSortedEqual(await adapter.query(projectQuery()), [p1._raw, p2._raw])
```
radex: can you check that the same call the second time will return ids? you're only testing caching after insert works, not that caching after query works

src/adapters/__tests__/commonTests.js
```diff
@@ -112,6 +114,72 @@ export default () => [
...
+      adapter = adapter.testClone()
+
+      const s2 = makeMockTask({ id: 'id2', text1: 'baz', order: 2 })
+      await adapter.batch([['create', s2]])
+
+      // returns cached ID after create
+      expect(await adapter.find('tasks', 'id2')).toBe('id2')
+
+      // returns raw if not cached for a different table
+      expect(await adapter.find('projects', 'id2')).toEqual(p2._raw)
```
radex: can you check that the same call the second time will return ids? you're only testing caching after insert works, not that caching after find works

src/adapters/lokijs/worker/executor.js
```diff
@@ -56,8 +56,36 @@ export default class LokiExecutor {
     await this._migrateIfNeeded()
   }
 
+  isCached(table: TableName<any>, id: RecordId): boolean {
+    const cachedSet = this.cachedRecords.get(table)
+    if (!cachedSet) {
+      return false
+    }
+
+    return cachedSet.has(id)
```
radex: i think we can nicely inline to:

```
return cachedSet ? cachedSet.has(id) : false
```

src/adapters/lokijs/worker/executor.js
```diff
@@ -56,8 +56,36 @@ export default class LokiExecutor {
...
+    cachedSet.add(id)
+  }
+
+  removeFromCache(table: TableName<any>, id: RecordId): void {
+    const cachedSet = this.cachedRecords.get(table)
+    if (!cachedSet) {
+      return
+    }
+
+    cachedSet.delete(id)
```
radex: instead of early return, do the delete inside if

src/adapters/lokijs/worker/executor.js
```diff
@@ -56,8 +56,36 @@ export default class LokiExecutor {
...
+  }
+
+  markAsCached(table: TableName<any>, id: RecordId): void {
+    let cachedSet = this.cachedRecords.get(table)
+    if (!cachedSet) {
+      cachedSet = new Set()
+      this.cachedRecords.set(table, cachedSet)
+    }
+
+    cachedSet.add(id)
```
radex: let's do it without local mutability:

cachedSet = xxxx

if ( cachedSet) {
cachedSet.add(x)
} else {
cachedRecords.set(table, new Set(x))

ahmadbaraka:
> @rozPierog take a look please ;) — could we avoid mutability in the Kotlin code similar to swift/js?

Fixed

rozPierog:
@ahmadbaraka can you fix conflict in `CHANGELOG`? Thanks!

ahmadbaraka:
@rozPierog I've resolved the conflict and done a rebase

radex:
Cool! I'll merge as soon as the CI status turns green!

@ahmadbaraka Do y'all want some Watermelon stickers? If you do, send me an email and I'll ship you some!

ahmadbaraka:
@radex I've sent you an email.

Is the failed build caused somehow by the changes in this PR?

radex:
Merging. Thanks @ahmadbaraka — nice work!

Kabangi:
When is this expected to be available on NPM?

radex:
@Kabangi after New Year — I don't want to break something while I'm away from work. You can just use @ahmadbaraka's fork for now — you'll find the instructions to build production package in Contributing.md

Issue #103, opened 2018-10-21
Raw Queries


developerium:
Hi 

Is there anyway we can run custom queries on database?

Is there anyway we can access database directly?

bsr203:
I used to be able to connect to the sqlite database through https://github.com/facebook/stetho

you can see the instructions at https://facebook.github.io/react-native/docs/debugging#debugging-with-stetho-http-facebookgithubio-stetho-on-android

But, not sure what changed, I cannot see my WatermelonDB database though `stetho` anymore, but other sqlite dbs. I like to know how people normally does it as well.

radex:
> Is there anyway we can run custom queries on database?

Not currently. What exactly are you trying to achieve?

MrTinyByte:
I want to make a report on several tables, a sum report on multiple joined tables.

radex:
https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/sqlite/index.js#L162

this is the code that executes the query. The query will pass through `encodeQuery` to generate SQL. You could maybe extend `Query` or make a new `RawQuery` object or something like that to be able to pass raw SQL. Contributions welcome.

If you want a quick & dirty hack (but I warn you that it's very dirty, might not work in the future or at all, and you don't have observability, and it might be unsafe):

```js
import { NativeModules } from 'react-native'
import { sanitizeQueryResult } from '@nozbe/watermelondb/adapters/common'

const tag = database.adapter._tag
const bridge = NativeModules.DatabaseBridge

const dirtyRecords = await bridge.query(tag, 'select from ...')
const rawRecords = sanitizeQueryResult(dirtyRecords, appSchema.tables['name_of_table'])

const records = database.collections.get('name_of_table')._cache.recordsFromQueryResult(rawRecords)
```

does this help?

radex:
@MrTinyByte hey, still interested in contributing to this? Did you try out the hack I described above?

servocoder:
I was looking a solution for SQL `order` and `limit` support, and found the explanations why it's not an easy thing considering observables, e.g. #15. While this is true for observables there are cases when you just need to fetch some DB entity out of the observables context. My particular case is that I need to find first entity sorted by timestamp with `fetch()` method. If we can't get it with `{Q}` then raw queries would become a solution.

Thanks to @zobeirhamid we have PR #199 and, if it's ok, I would vote for it to be merged.

@radex take a look this PR please.

sursma:
[...]

PR #199, opened 2019-01-09
Adds rawQuery functionality

zobeirhamid:
Adds rawQuery functionaility, but needs Flow coverage.

Usage:
```js
const tableNameCollection = database.collections.get('table_name')
tableNameCollection.unsafeFetchRecordsWithSQL('select * from ...')

database.adapter.unsafeSqlQuery('select * from ...', 'table_name')
```
Issue: #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
+- Raw Query
 
 ### ⚠️ Breaking
 
--- a/docs/Query.html
+++ b/docs/Query.html
@@ -4,7 +4,7 @@
         <!-- Book generated using mdBook -->
         <meta charset="UTF-8">
         <title>Querying - WatermelonDB documentation</title>
-        
+
 
         <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
         <meta name="description" content="">
@@ -28,9 +28,9 @@
         <link rel="stylesheet" href="ayu-highlight.css">
 
         <!-- Custom theme stylesheets -->
-        
 
-        
+
+
     </head>
     <body class="light">
         <!-- Provide site root to javascript -->
@@ -86,7 +86,7 @@
         <div id="page-wrapper" class="page-wrapper">
 
             <div class="page">
-                
+
                 <div id="menu-bar" class="menu-bar">
                     <div id="menu-bar-sticky-container">
                         <div class="left-buttons">
@@ -103,11 +103,11 @@
                                 <li role="none"><button role="menuitem" class="theme" id="navy">Navy</button></li>
                                 <li role="none"><button role="menuitem" class="theme" id="ayu">Ayu</button></li>
                             </ul>
-                            
+
                             <button id="search-toggle" class="icon-button" type="button" title="Search. (Shortkey: s)" aria-label="Toggle Searchbar" aria-expanded="false" aria-keyshortcuts="S" aria-controls="searchbar">
                                 <i class="fa fa-search"></i>
                             </button>
-                            
+
                         </div>
 
                         <h1 class="menu-title">WatermelonDB documentation</h1>
@@ -116,12 +116,12 @@
                             <a href="print.html" title="Print this book" aria-label="Print this book">
                                 <i id="print-button" class="fa fa-print"></i>
                             </a>
-                            
+
                         </div>
                     </div>
                 </div>
 
-                
+
                 <div id="search-wrapper" class="hidden">
                     <form id="searchbar-outer" class="searchbar-outer">
                         <input type="search" name="search" id="searchbar" name="searchbar" placeholder="Search this book ..." aria-controls="searchresults-outer" aria-describedby="searchresults-header">
@@ -132,7 +132,7 @@
                         </ul>
                     </div>
                 </div>
-                
+
 
                 <!-- Apply ARIA attributes after the sidebar and the sidebar toggle button are added to the DOM -->
                 <script type="text/javascript">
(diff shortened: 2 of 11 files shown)


radex:
@zobeirhamid Thanks for the PR! This is a good start, but there's a couple of things we need before I can merge this:

- [ ] Flow declarations in code
- [ ] Tests — see https://github.com/Nozbe/WatermelonDB/issues/112 for pointers as to where to add tests
- [ ] Docs update

zobeirhamid:
@radex Thanks for the directions. Planning to work on it when I have some free time, probably in two weeks.

servocoder:
@zobeirhamid any update on this?

wkirby:
Any update here? This feature is pretty necessary to use WatermelonDB in any serious capacity.

radex:
@wkirby You're welcome to contribute if you want to pick up where @zobeirhamid left off and submit a new PR :)

zobeirhamid:
I updated the requested tasks. I only had one problem writing the tests since the raw query functionality is only implemented for SQLite and not for Loki, but since it's a new method I had to add it for the LokiJS Adapter too, but left the implementation open.

radex:
Awesome!

> I only had one problem writing the tests since the raw query functionality is only implemented for SQLite and not for Loki,

@zobeirhamid look at this - this is a way of determining whether you're running tests in web or native mode

https://github.com/Nozbe/WatermelonDB/blob/7b1546ad4219194cf4f3a03f0ad3a56761ef2a31/src/adapters/__tests__/commonTests.js#L777

radex:
@zobeirhamid there are still some CI issues :/  take a look at Travis CI details

zobeirhamid:
@radex I fixed the previous issues but still get a fail even I don't see any obvious errors.

radex:
looks like this test: https://github.com/Nozbe/WatermelonDB/pull/199/files#diff-18ce2a16e966068b4dc9aed56cd01ff8R233 gets stuck and never completes.

have you tried running native tests locally?

zobeirhamid:
@radex On my machine, all tests are passing. I rebased the fork with the repo and tried running the tests again and they are still passing. I, to be honest, don't know what is going wrong.

radex:
Sure, I'll try to take a look at it this friday. Thanks for working on this @zobeirhamid !

jonnywhite:
Hey @radex 
Do you know if this is still being looked at? We are really keen for this functionality too.
Thanks!

radex:
@jonnywhite Hey! Apologies! Probably in two weeks time, after I get back to work… I know it's taking too much time, sorry about that.

joshuakelly:
@zobeirhamid Are you also able to do a `SUM` on a column with this PR? It's only returning a single record for me when i run the query.

zobeirhamid:
@joshuakelly SUM is suppose to only return a single record, in fact it should return the SUM of a column for all rows.
https://www.w3schools.com/sql/trysql.asp?filename=trysql_select_sum

radex:
Thanks for your work on this @zobeirhamid - and apologies for late replies.

This is super useful, but I have some concerns:

- why does RawQuery extend Query if it doesn't actually support most of its methods?
- what about web/Loki? It's OK if there's no support for it in the first version, but the `RawQuery` type shouldn't be geared towards `sql` only but should be more general - more like `query: AdapterQuery` 
- it seems that RawQuery is not… raw enough! I mean, it only supports queries that return a list of records. But what about queries that return, say, a number, or a list of numbers? or don't return anything (not query but raw execution)? or some other crazy SQL options I can't even think of...

I don't expect you to have all the answers :) And I'm all for iterative development approach and taking one step at a time. But I don't like doing breaking changes, and we don't want to end up with an API that sucks in the long term, so there's two options:

1. Either we discuss and figure out an API that we'll be able to extend in the future to take care of all the concerns above without breaking anything
2. Or we mark this API in such a way that it's clear it's limited in functionality and might change in the future. For example, instead of `collection.rawQuery('sql'): RawQuery; rawQuery.fetch()`, we could do `collection.unsafeFetchRecordsWithSQL(sql: SQL): Promise<Record[]>`.

Option 1 is better, and option 2 means you can just rename a few things in your PR and we can merge soon :)

zobeirhamid:
@radex 
Thanks for the feedback. I agree with your points. I'm gonna give some thoughts about a new API design in the next couple of days. I would have just need all the supported output types, as you listed: records, numbers, and list of numbers. Are there anymore, or is there any particular place to look for the supported output types?

radex:
> Are there anymore, or is there any particular place to look for the supported output types?

You mean in SQL in general? TBH I don't know.

Currently 🍉 standard queries (query, find, count) only fetch records, lists of records, and integers (counts of records) — but for raw queries, we don't want to be limited to that...

zobeirhamid:
@radex Me neither :D
Alright, so basically you imagine to be able to use sqlite through watermelon? If that's the case, wouldn't that imply that code won't be transportable between web and mobile anymore?
Personally, I don't like that approach, since if you really want, you can use react-native-sqlite-storage or similar in combination with watermelon, to use raw instructions (I tried it, it works). I would prefer, a semi-raw approach where we be open to raw input, but would limit the output format with that we would allow more complex queries and maintain the transportability of the code.

radex:
> I would prefer, a semi-raw approach where we be open to raw input, but would limit the output format with that we would allow more complex queries and maintain the transportability of the code.

What do you mean exactly?

I’m not quite sure what sort of output format can you even get from what sort of SQLite query that would affect cross-platform ness more than the fact itself that you’re using SQL. (I think it’s natural that there would also be raw queries for LokiJS). But I think you can do queries that return, say, a list of numbers (doing statistics and such on the data) that would be useful

zobeirhamid:
It could be that I have an error in my thought process, but what I imagined is if you execute a raw select query, then you should receive your class objects instead of just raw objects, since if you want really raw operation functionality, you can just use another library.
Nevertheless, I understand if you watermelon doesn't want to depend on another library for real raw operations, if that's the case I can also do a contribution with real raw query.

radex:
> but what I imagined is if you execute a raw select query, then you should receive your class objects

yes, of course, but again -- with sql you can also fetch numbers...

so how about we just give up for now with preparing an extensible API and just leave `Collection.fetchRawQuery(sql: SQL): Promise<Record[]>` ? I think other open source contributors which have other needs than just fetching record lists by SQL might have good ideas as for how to go about this

vdlindenmark:
What's the status of this feature?

radex:
@vdlindenmark waiting for response from PR author. You can read through the discussion - there isn't a lot of work left until this is mergeable, though some things need to be reworked. You are more than welcome to take over this PR and re-submit it if you want to finish it.

zobeirhamid:
Sorry for the wait; I totally forgot about this pull request, but here it is.
Renamed the methods and etc.
Wrote the test for the adapter method, but couldn't figure out how to write the test for the collection method since it mocks the LokiJS Adapter, but since it's an unsafe method that shouldn't be too big of a deal.

yanielbf:
@zobeirhamid hi, I'am using your fork. I builded and installed in my project and work good, but only at debug, not when generate the release . Help me please. Greetings

radex:
thanks @zobeirhamid ! I'm closing this PR in favor of: https://github.com/Nozbe/WatermelonDB/pull/557 — it's branching off from your branch, but I had to make some changes, and couldn't push to your fork. Expect it to be merged and released next week

code-brewer:
Failed to get record numbers.  Anythie wrong with my code? 
    The query return nothing, nor 0, nor undefined, ...

```
let num = await db.adapter.unsafeSqlQuery('posts', 'select count(1) from posts where 1=1;');
// also try unsafeSqlQuery('posts', 'select count(1) from posts;');   got the same result.
console.log("getDocsNumsBySql: return " + num); 
```

radex:
@code-brewer unsafeSqlQuery expects a list of records… sadly other types of queries, such as a single number are not supported. You're welcome to send a pull request improving the API.

code-brewer:
@radex  I am sorry I can't give PR because of little spare time. But I can post issues here. I did some pressure tests this week and found memory strange and unreasonably high on iOS while android seems sane.

PR #201, opened 2019-01-10
a little more explanation for Q.like

anarkafkas:


radex:
Thanks @anarkafkas — take a look at rendered markdown. Looks like there are some `\` that shouldn't be there. Also, the second example is incorrect

anarkafkas:
Sorry I'm relatively new to this. Not quite familiar with js regex. I think my 2nd example had 2 extra `\`s. I can confirm, in our app we query all users whose usernames match (not necessarily start with) the pattern "jas"  as such:

```js
usersCollection.query(
  Q.where("username", Q.like(`%${Q.sanitizeLikeString("jas")}%`)
)
```

Also the first inline code didn't render properly, so I moved it onto a new line but again not quite sure about the `\`. If I don't touch it and just move to a new line it is

```js
Q.like(\`%${Q.sanitizeLikeString(userInput)}%\`)
```

but I get "Invalid Character" error for this one.

rkrajewski:
@anarkafkas Both examples shouldn't have backslashes:

```js
Q.like(`%${Q.sanitizeLikeString(userInput)}%`)
```

and the second one shouldn't have leading `%` if searching string **starts with** `jas`:

```js
Q.where("username", Q.like(`${Q.sanitizeLikeString("jas")}%`)
```

PR #209, opened 2019-01-18
Gradle update

rozPierog:
2018 come to an end, so it is high time to update `compile` to `implementation` in gradle files. 
Also I've updated Kotlin, Gradle, and moved from `android.support` to `androidx`

### DIFF
--- a/examples/native/android/app/build.gradle
+++ b/examples/native/android/app/build.gradle
@@ -97,7 +97,6 @@ def enableProguardInReleaseBuilds = false
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
     buildToolsVersion rootProject.ext.buildToolsVersion
-
     defaultConfig {
         applicationId "com.app"
         minSdkVersion rootProject.ext.minSdkVersion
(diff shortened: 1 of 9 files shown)


radex:
Don't we need to bump package.json react-native peer dependency to be at least 0.57 then?

rozPierog:
@radex I've run test app locally and everything was alright so I haven't given it a second thought, but I'm pretty sure that everything _should_ work with `"react-native": "^0.56.0"` obviously it would hurt to bump it up to `0.57`. Should I do it in this PR?

radex:
> I'm pretty sure that everything should work with "react-native": "^0.56.0" 

wasn't there the thing that on 0.56, unless your project is specifically configured to support gradle 3, it will fail to compile?

rozPierog:
Oh yeah right, I was under impression that change was made in 0.56 but in reality it was in 0.57 as you say. Anyway, question still stands, should I bump `peerDependecy` `react-native` to 0.57 or not? If not, in coming days/weeks/months Android part of library may just fail to compile for someone who is updating their Gradle.

radex:
Yes, we need to bump peerDependency. People can still choose to use Watermelon on 0.56 (ignoring the peer dependency warning) if they know what they're doing, but it won't work by default, so let's bump it

rozPierog:
@radex Done!

radex:
@rozPierog could you also please add a note to CHANGELOG that you've bumped gradle and RN dependency, so it might break build?

rozPierog:
@radex I didn't mention changes inside `example` and `test` apps because I feel like this changelog isn't the right place for this. Is this okay?

PR #223, opened 2019-02-02
Fix crash on android < 5.0

Kabangi:
For some reason, this line here `delete from sqlite_master` causes database corruption in `android < 5.0` and the app would always crash on startup.

In this PR, I only delete the necessary metadata from `sqlite_master` instead of everything.

#### Test Plan 
1. Build the example RN repo post this PR 
2. Observe the app would crash with database corruption error on log cat
3. Rebuild with this PR
4. Observe everything turns out to work great.


Kabangi:
@radex Thanks for the reminder I intended to do it but just completely forgot about it.

radex:
@Kabangi would you mind adding a note to the CHANGELOG.md file?

Kabangi:
@radex Done

PR #227, opened 2019-02-04
Add simple support of ejected expo projects

ieschalier:
Allow WatermelonDB to support **ejected** expo projects. 🤘


radex:
Thanks @ieschalier ! Would you mind adding a quick note to CHANGELOG about this fix? :)

PR #233, opened 2019-02-06
[iOS] Allow passing path instead of name to SQLiteAdapter

radex:
- Allow passing path instead of name to SQLiteAdapter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,15 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
-- Changed `compile` to `implementation` in Library Gradle file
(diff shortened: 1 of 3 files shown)


avinashlng1080:
@radex does this feature applies only to iOS ? Can I also pass a path to the document directory on Android?  I tried and it isn't working.  I'm getting the error : 

```
[WatermelonDB] Uh-oh. Database failed to load, we're in big trouble. This might happen if you didn't set up native code correctly (iOS, Android), or if you didn't recompile native app after WatermelonDB update. It might also mean that IndexedDB or SQLite refused to open. [Error: File file:///data/user/0/com.mattermost.rnbeta/files/databases/default.db/default.db.db contains a path separator]
```

PR #240, opened 2019-02-07
Update Installation.md

cesargdm:
Added specified module to link at react-native link, if no module is specified all modules are linked, and there area some RN libraries that explicitly request to not link the library thru RN

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -31,7 +31,7 @@ yarn add @nozbe/watermelondb
     **Automatically**
 
     ```bash
-    react-native link
+    react-native link @nozbe/watermelondb
     ```
 
     **Or manually**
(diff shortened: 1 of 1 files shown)


radex:
Nice! didn't know you can do that.  Thanks @cesargdm . While you're at it, could you re-order the iOS installation steps, so that it FIRST asks you to add Swift support, and THEN to do `react-native link`? Many people just stop paying attention after the linking part and then file issues saying their app doesn't compile

cesargdm:
@radex Done, removed the "If you get linker errors when building, you need to...", because I think is clearer just "Add Swift support to the project", what do you think?

radex:
Agreed! Thanks @cesargdm :)

PR #244, opened 2019-02-12
Fix @date getter

radex:
`@date` field now returns a Jan 1, 1970 date instead of `null` if the field's raw value is `0`.

### DIFF
--- a/src/decorators/date/index.js
+++ b/src/decorators/date/index.js
@@ -24,7 +24,7 @@ const dateDecorator = makeDecorator(
       enumerable: true,
       get(): ?Date {
         const rawValue = this._getRaw(columnName)
-        return rawValue ? new Date(rawValue) : null
+        return typeof rawValue === 'number' ? new Date(rawValue) : null
       },
       set(date: ?Date): void {
         const rawValue = date ? +new Date(date) : null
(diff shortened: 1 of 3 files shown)


radex:
@rkrajewski CR please

src/decorators/date/index.js
```diff
@@ -24,7 +24,7 @@ const dateDecorator = makeDecorator(
       enumerable: true,
       get(): ?Date {
         const rawValue = this._getRaw(columnName)
-        return rawValue ? new Date(rawValue) : null
+        return typeof rawValue === 'number' ? new Date(rawValue) : null
```
rkrajewski: Just wondering should there be any warning, if raw value of date field will be `NaN` somehow. Since `typeof NaN` is `number`, this will produce Invalid Date. WDYT?

radex: Good catch! Thankfully, there's nothing to worry about:

https://github.com/Nozbe/WatermelonDB/blob/6100ae2a9208ebd3cb2ffc5624fa7072e631d4ee/src/RawRecord/index.js#L59

the raw values go through this sanitizer, that will ignore invalid numbers like NaN of Infinity

PR #246, opened 2019-02-13
Sync protect against clear database

radex:
- ⚠️ Potentially breaking: `Database.unsafeResetDatabase()` now requires that you run it inside an Action
- [Sync] Fixed an issue where synchronization would continue running despite `unsafeResetDatabase` being called
- Actions are now aborted when `unsafeResetDatabase()` is called, making reseting database a little bit safer
- Clean up, document, update tests


rozPierog:
@radex annnd another one 😶

PR #261, opened 2019-02-21
Update Actions.md

cesargdm:
As speciefied in https://eslint.org/docs/rules/no-return-await `Inside an async function, return await is seldom useful. Since the return value of an async function is always wrapped in Promise.resolve, return await doesn’t actually do anything except add extra time before the overarching Promise resolves or rejects. The only valid exception is if return await is used in a try/catch statement to catch errors from another Promise-based function.`

Or maybe no return at all?

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -14,8 +14,8 @@ import { action } from '@nozbe/watermelondb/decorators'
 class Post extends Model {
   // ...
 
-  @action async addComment(body, author) {
-    return await this.collections.get('comments').create(comment => {
+  @action addComment(body, author) {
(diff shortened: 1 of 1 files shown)


radex:
Thanks @cesargdm, but I disagree with this recommendation, and here's why:

- Your method starts as a simple one, with just a single Promise return. There's no point for async and no point in await
- But then you realize you need to modify your method and add your action. You add "await thisAndThat", and it doesn't work. Ahh! I forgot `async`. Now it all works!
- Except it's wrong, because you didn't add `await` before the first asynchronous operation!

cesargdm:
Yup, in terms of explanation I agree the current way is clearer.

PR #289, opened 2019-03-19
Sync log

radex:
- Adds basic logging capability to `sync`. This can be easily extended in the future

### DIFF
--- a/docs/Advanced/Sync.md
+++ b/docs/Advanced/Sync.md
@@ -9,8 +9,6 @@ Note that Watermelon is only a local database — you need to **bring your own b
 
 ## Using `synchronize()`
 
-> ⚠️ The `synchronize()` function is not yet formally released (see changelog & pull requests for more details) and is currently experimental. There are extensive tests for this sync adapter, but it has not yet been battle tested. It's also not yet fully optimized, so you might experience less than ideal performance during sync.
-
 Using Watermelon sync looks roughly like this:
 
 ```js
@@ -42,7 +40,7 @@ async function mySync() {
 
 ```
 
-You need to pass two functions, `pullChanges` and `pushChanges` that can talk to your backend in a compatible way.
+You need to pass two functions, `pullChanges` and `pushChanges` that can talk to your backend in a compatible way (explained later).
 
 **⚠️ Note about a React Native / UglifyES bug**. When you import Watermelon Sync, your app might fail to compile in release mode. To fix this, configure Metro bundler to use Terser instead of UglifyES. Run:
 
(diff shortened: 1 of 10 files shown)


docs/Advanced/Sync.md
```diff
@@ -170,6 +168,20 @@ Synchronization is serious business! It's very easy to make mistakes that will c
...
   - If there's something wrong with the data format, prefer to "fix" the data if possible instead of failing sync. You don't want the user to have an unsyncable app because of a mistake caused by a bug 5 versions ago.
   - As with any API, remember to check permissions to create/modify records, make sure you version your API together with local Schema versioning, and all other standard stuff!
+- **Adding logging to your sync**
+  - You can add basic sync logs to the sync process by passing an empty object to `synchronize()`. Sync will then mutate the object, populating it with diagnostic information (start/finish time, resolved conflicts, and more):
+
+    ```js
+    const log = {}
+    await synchronize({
+      database,
+      log,
```
kokusGr: Hmm wouldn't it be better if user only passes a flag `log: true` and `synchronize` returns `log` object without mutating?

src/sync/impl/helpers.js
```diff
@@ -54,10 +55,29 @@ export function prepareCreateFromRaw<T: Model>(collection: Collection<T>, dirtyR
...
-export function prepareUpdateFromRaw<T: Model>(record: T, updatedDirtyRaw: DirtyRaw): T {
+export function prepareUpdateFromRaw<T: Model>(
+  record: T,
+  updatedDirtyRaw: DirtyRaw,
+  log: SyncLog,
+): T {
+  // Note COPY for log - only if needed
+  const logConflict = !!record._raw._changed
+  const logLocal = logConflict ? { ...record._raw } : {}
+  const logRemote = logConflict ? { ...updatedDirtyRaw } : {}
```
kokusGr: Shouldn't this be done only if user wants logs?

docs/Advanced/Sync.md
radex: BUSTED! 🚓 

You're right, it would be cleaner — I was being lazy ;P Still… there is an issue. `synchronize()` will throw an error (i.e. reject the returned promise) if sync fails — you won't get a log then. So we could either resolve with a log or catch and re-throw all errors inside synchronize by wrapping them in a SyncError and attach a log there… Then a client would have to also handle both cases…

was supposed to be cleaner but… gets a lot uglier. do you think it's worth it?

src/sync/impl/helpers.js
radex: sure, I can do that

docs/Advanced/Sync.md
kokusGr: ugh I guess you're right... On the other hand I think that people interested in logs will also catch sync errors anyway but for now let's leave it as it is.

src/sync/impl/helpers.js
radex: fixed

radex:
also added some flow coverage / etc improvements

kokusGr:
LGTM :)

PR #294, opened 2019-03-23
Enable batch delete

esbenp:
This commit enables batch deletes either through prepareMarkAsDeleted or prepareDestroyPermanently

What do you think of this implementation? @radex I do not know the code base or rxjs well enough to see any potential pitfalls. I have only done very limited testing of this

**TODO**
- [ ] Tests
- [ ] Documentation
- [ ] Fix types


radex:
@esbenp Awesome! This is good stuff, I was really hoping someone from the community would try and tackle this :) I will take a closer look later in the week (probably Friday), but the general approach is sound.

src/Model/index.js
```diff
@@ -118,6 +120,28 @@ export default class Model {
     return this
   }
 
+  prepareMarkAsDeleted(): this {
+    invariant(this._isCommitted, `Cannot mark an uncomitted record as deleted`)
```
radex: I *think* we should also protect againt deleting a record with pending update

src/Model/index.js
```diff
@@ -118,6 +120,28 @@ export default class Model {
...
+    this._raw._status = 'deleted'
+    this._hasPendingDelete = 'destroy'
+    this._isEditing = false
+
+    return this
+  }
+
   // Marks this record as deleted (will be permanently deleted after sync)
   // Note: Use this only with Sync
   async markAsDeleted(): Promise<void> {
```
radex: I think most of markAsDeleted/destroyPermanently methods should be replaced with a call to database.batch(prepareXxxx) — similar to how model.update mostly just does database.batch(model.prepareUpdate). This is to avoid duplicate code and two code paths doing essentially the same thing. There's also some support code in Collection we could then delete

src/Model/index.js
esbenp: I thought about this as well. In the end my conclusion was deletion should just take precedence, i.e. if a user tries to delete an updated record - does it then matter it was not synced? It should still delete, right?

src/Model/index.js
esbenp: Roger that. I will try to put in some finishing touches that accounts for this

src/Model/index.js
radex: Right, but "pending update" means that someone called `prepareUpdate()` synchronously (so just before) calling `prepareMarkAsDeleted()` (not that it's pending sync or something like that). And I can't think of a good reason to batch both an update and a delete for a record — so if someone does that, it's almost surely a mistake. We can always relax this invariant, but I just fear some weird edge case showing up down the line without any test covering it

src/Model/index.js
radex: awesome 👍

src/Model/index.js
esbenp: I would suggest issuing a warning in dev mode, but still doing the delete. What do you think?

radex: I suggest doing an error. My reasoning: it's easier to relax this if someone complains that this breaks their workflow, then to tighten it and break people's apps if someone relies on this without a good reason

radex:
@esbenp Hey, what's the status of the PR? I don't think there's too much work left until we can merge this :)

esbenp:
It's been a while since I took a look at it since I have been on holidays. As far as I can tell the only thing I was missing was some sort of integration test of the deletes, which I just sort-of added in your own big batch test and then removed the old delete test in collection tests. What do you think?

radex:
@esbenp Looks great! I'll do a proper code review on Friday and then hopefully we can merge.

There's some CI fails: https://travis-ci.com/Nozbe/WatermelonDB/jobs/193601055 — you can run it locally with `yarn c:check`

radex:
@esbenp Would you mind also adding a sentence or two about this to the documentation?  https://github.com/Nozbe/WatermelonDB/blob/master/docs/Actions.md#batch-updates

esbenp:
@radex feel free to change any wordings as you see fit

esbenp:
hmmm.. not really sure what happened with the commits here

docs/Actions.md
```diff
@@ -53,7 +53,7 @@ await comment.markAsSpam()
 
 ## Batch updates
 
-Whenever you make more than one change (create or update records) in an action, you should **batch them**.
+Whenever you make more than one change (create, delete or update records) in an action, you should **batch them**.
```
radex: could you just add a line or two of explanation later in this document? It's explained that you can pass prepared operations but nowhere is it explained that you can specifically use `prepareDestroyPermanently` and the other one...

src/Database/index.js
```diff
@@ -55,11 +55,18 @@ export default class Database {
...
       )
 
-      if (record._hasPendingUpdate) {
+      // Deletes take presedence over updates
+      if (record._hasPendingDelete !== false) {
+        if (record._hasPendingDelete === 'destroy') {
+          return ops.concat([['destroyPermanently', record]])
+        }
+        
+        return ops.concat([['markAsDeleted', record]])
```
radex: you can replace the 5 lines above with just one (with a ternary record._hasPendingDelete === 'destroy' ? 'destroy...' : 'markAsDeleted') ?

src/Database/index.js
```diff
@@ -70,9 +77,23 @@ export default class Database {
 
     const sortedOperations: { collection: Collection<*>, operations: CollectionChangeSet<*> }[] = []
     operations.forEach(([type, record]) => {
+      const operationTypeToCollectionChangeType = input => {
```
radex: could you move this function to the file top-level (outside the Database class)? or maybe Database/helpers.js if this exists? as far as i can see the only argument is type, so it's better for clarity and performance not to create this function every time .batch() is called.

src/Collection/test.js
```diff
@@ -206,24 +206,4 @@ describe('creating new records', () => {
...
-
-    const m1 = new MockTask(collection, { id: 'm1' })
-    collection._cache.add(m1)
-
-    await collection._destroyPermanently(m1)
-
-    // Check database delete, cache delete, observers update
-    expect(adapter.batch).toHaveBeenCalledTimes(1)
-    expect(adapter.batch).toHaveBeenCalledWith([['destroyPermanently', m1]])
-    expect(collection._cache.get('m1')).toBe(undefined)
```
radex: this looks like something that's missing from your refactored test

src/Model/index.js
```diff
@@ -125,15 +127,38 @@ export default class Model {
...
     this.collection.database._ensureInAction(
       `Model.markAsDeleted() can only be called from inside of an Action. See docs for more details.`,
     )
-    invariant(this._isCommitted, `Cannot mark as deleted uncommitted record`)
-    this._raw._status = 'deleted'
-    await this.collection._markAsDeleted(this)
+    this.prepareMarkAsDeleted()
+    await this.collection.database.batch(this)
```
radex: you can do it in one line, batch(this.prepareMarkAsDeleted()) :)

src/Model/index.js
```diff
@@ -142,8 +167,8 @@ export default class Model {
     this.collection.database._ensureInAction(
       `Model.destroyPermanently() can only be called from inside of an Action. See docs for more details.`,
     )
-    invariant(this._isCommitted, `Cannot destroy uncommitted record`)
-    await this.collection._destroyPermanently(this)
+    this.prepareDestroyPermanently()
+    await this.collection.database.batch(this)
```
radex: same here

src/Collection/test.js
esbenp: Could you expand a bit on what test you want? Is it not covered by https://github.com/Nozbe/WatermelonDB/pull/294/files#diff-1c7bf3678f7379f801755def5d7c2facR156 ?

src/Database/helpers.d.ts
```diff
@@ -0,0 +1,6 @@
+import { BatchOperationType } from "@nozbe/watermelondb/adapters/type";
+
+declare module '@nozbe/watermelondb/Database/helpers' {
```
esbenp: I could use some guidance her @radex - have never written ts before - is this correct?

esbenp:
@radex do you need any further changes?

src/Database/helpers.d.ts
radex: TBH I don't know :) We don't use TypeScript internally, 🍉's TS support is entirely community-led

radex:
Apologies for ignoring this for a while. I'm now back from conferences & vacation :)

This looks good! Thank you so much @esbenp :) There's more work to be done - right now we can only batch-delete a single record, and it would be great to be able to efficiently delete whole trees of records, but that requires some extra work. Similarly with adapting Sync to use this. In case you'd be interested in helping out with this - let me know and I'll give you pointers as to how to implement this.

PS. Would you like some WatermelonDB stickers? If so, please email me with your address and I'll send some to you.

esbenp:
No worries @radex 

Thanks for merge and welcome back :-)

src/Database/helpers.d.ts
Bessonov: TBH “But why not TypeScript”? :D https://sobolevn.me/2019/03/from-flow-to-typescript

I hope to look at it this week.

PR #297, opened 2019-03-25
Fix (?) compile for schemes with whitespace

radex:


vdlindenmark:
See #298 for a better fix

PR #302, opened 2019-03-27
Fix runtime crash when built with Xcode 10.2 (Swift 5 runtime)

radex:


radex:
@rkrajewski CR please - merge if Purple with this fix works for you

radex:
@rkrajewski would you mind taking a look on Monday? It would be good to release this, as droves of people will have their Xcode updated and won't know what to do about it

radex:
@kilbot Awesome! I'll still have to tweak this to use the new API

radex:
OK, done. Let's wait to see if Travis has no problem with this

PR #308, opened 2019-04-03
Sync batch per collection

radex:
- [Sync] Adds a temporary/experimental `_unsafeBatchPerCollection: true` flag to `synchronize()`. This
      causes server changes to be committed to database in multiple batches, and not one. This is NOT preferred
      for reliability and performance reasons, but it works around a memory issue that might cause your app
      to crash on very large syncs (>20,000 records). Use this only if necessary. Note that this option
      might be removed at any time if a better solution is found.


radex:
@kokusGr CR please :)

radex:
> Looks ok :) But I'm hoping we find better solution soon 😰

I think we'll leave it as is for a couple of months. When React Native TurboModules hit, maybe we can do without any hacks. If not, we can do adaptive batching maybe (split automatically based on number of records)? Or refactor the code to be… well… less FP-y (do less copying) — if really, really, really necessary

PR #319, opened 2019-04-13
Added npm install instructions along with yarn

### LINKED ISSUE #299: Add npm in install instructions along with yarn
I have observed that in the installation instructions, there is a preference of `yarn` over `npm` for the purpose of installation and setup of `watermelonDB`.
While in most cases this is preferred, there are many users who prefer to only use npm as their default package manager in an attempt to aid simplicity.

Shouldn't it be better to add support for npm as well, I pre-assume that npm should also be supported out of the box and this shouldn't be involving anything other than documentation files.
I would also love to work on this as my first contribution here 🥇

co16353sidak:
I have attempted to add npm install instructions along with the yarn install instructions in the guide wherever applicable.
I have checked the usage of the commands by using them on the cli and I feel that more or less they will work the same as yarn. 
Everything was done as discussed in #299 

#trivial 

(Fixes #299)

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -6,6 +6,12 @@ First, add Watermelon to your project:
 yarn add @nozbe/watermelondb
 yarn add @nozbe/with-observables
 ```
+or alternatively if you prefer npm:
+
+```npm
+npm install @nozbe/watermelondb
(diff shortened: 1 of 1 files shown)


radex:
Thanks!

PR #323, opened 2019-04-15
[Database] Allow falsy values to batch()

radex:
[Database] You can now pass falsy values to `Database.batch(...)` (false, null, undefined). This is
     useful in keeping code clean when doing operations conditionally

### DIFF
--- a/src/Database/index.js
+++ b/src/Database/index.js
@@ -43,12 +43,17 @@ export default class Database {
 
   // Executes multiple prepared operations
   // (made with `collection.prepareCreate` and `record.prepareUpdate`)
-  async batch(...records: $ReadOnlyArray<Model>): Promise<void> {
+  // Note: falsy values (null, undefined, false) passed to batch are just ignored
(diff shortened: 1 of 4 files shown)


src/Database/index.js
```diff
@@ -43,24 +43,29 @@ export default class Database {
 
   // Executes multiple prepared operations
   // (made with `collection.prepareCreate` and `record.prepareUpdate`)
-  async batch(...records: $ReadOnlyArray<Model>): Promise<void> {
+  // Note: falsy values (null, undefined, false) passed to batch are just ignored
```
kokusGr: I think this also should be in docs/changelog :)

radex: done

kokusGr: I meant info that falsy values are ignored but info about `model.batch` is also good :)

radex: fixed on master https://github.com/Nozbe/WatermelonDB/commit/f8922d2ba11b29d7db744d2071feda6a55c7a292

kokusGr:
Ok, thanks! :)

Issue #7, opened 2018-09-06
Override entity IDs


sebastian-schlecht:
Hey,

we are currently evaluating this for our app. Since we try to implement our own sync engine on top, I'd be curious if it's possible to specify entity IDs whenever you create one.

Right now, if I try to set an ID in the builder function, I get the following error:

`Diagnostic error: Attempt to set new value on a property Model.prototype.id marked as @readonly`

Great work by the way, this looks very promising!

radex:
Yep, for sync you should use a lower-level API. I'll try to document this better sometime soon.

But in short, you can either use:

```js
record._setRaw('id', myId)
```

But if you're pulling a record from the server, and you have a simple JSON-like object with the record's data, you can also do:

```js
import { sanitizedRaw } from 'watermelondb/RawRecord'

collection.create(record => {
  record._raw = sanitizedRaw({ id: xxx, foo: x, bar: z }, collection.schema)
})
```

radex:
right, this doesn't work for `id`. You can do `record._raw.id = myId`, but this might be unsafe — tell me what exactly you're trying to do

Kabangi:
I want to use the same `id` am getting from my server as the record `id`. This will help me simplify detection of whether a record exists on the local cache using the withObservable HOC before attempting to pull it from the server.

radex:
so when exactly would you want to assign an ID to the record? In the sync code? Can you give me a larger snippet of code?

radex:
@Kabangi have you figured it out? you need something like this:

```js
const contactCollection = database.collections.get('contacts')
contactCollection.create(record => {
  record._raw = sanitizedRaw({ id: c.id, sid: c.id, username: c.username, .... }, contactCollection.schema)
})
```

Kabangi:
Yes, ended up using  ._raw.id=id, though am afraid in future there might be a breaking change in regard to this.

Kabangi:
@radex 
There is a gotcha with this solution `record._raw.id = myId`. That is, the ids from the server on all the models that you have needs to be unique otherwise you endup having this kind of errors

```
Record ID was sent over the bridge, but it's not cached
```

This is mainly because watermelon uses a Map with the `id` as the key hence if you have multiple models with the same id they will collide.

adam-aerobotics:
@Kabangi 
I just noticed this issue myself! We're setting all the ids to our own database ids, but it seems a few that are relations have the same ids and are throwing that same error.

[...]

PR #328, opened 2019-04-21
Allow custom id generator

sebastian-schlecht:
Minor refactor of `randomId` that allows users to provide a custom id generator function to WatermelonDB in order to create specific random id formats (e.g. if UUIDs are used in the backend)

Relates to #7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ All notable changes to this project will be documented in this file.
 - [TypeScript] 'Cannot use 'in' operator to search for 'initializer'; decorator fix
 
 ### Changes
-
+- [Database] You can now update the random id schema by importing `import { setGenerator } from '@nozbe/watermelondb/utils/common/randomId'` and then calling `setGenerator(newGenenerator)`. This allows WatermelonDB to create specific IDs for example if your backend uses UUIDs.
 - [Database] You can now pass falsy values to `Database.batch(...)` (false, null, undefined). This is
     useful in keeping code clean when doing operations conditionally. (Also works with `model.batch(...)`)
 - [Decorators]. You can now use `@action` on methods of any object that has a `database: Database`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Changes
+- [Database] You can now update the random id schema by importing `import { setGenerator } from '@nozbe/watermelondb/utils/common/randomId'` and then calling `setGenerator(newGenenerator)`. This allows WatermelonDB to create specific IDs for example if your backend uses UUIDs.
 
 ## 0.12.2 - 2019-04-19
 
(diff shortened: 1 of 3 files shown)


rkrajewski:
@sebastian-schlecht Can you consider to keep default export from randomId as function, that calls generator stored in module scope (with setter for new generator)? Something like:

```js
let generator = () => randomId()

export const setGenerator = _generator => {
  generator = _generator
}

export default () => generator()
```

- This approach won't break dependent code - since randomId can be exported outside of Watermelon - this is not so big deal, but still.
- It will require less changes in the Watermelon code - just an extra test.
- In current solution RandomId class is singleton anyway, and it's a bit more complicated solution. You can also avoid any checks from updateGenerator, just by overwriting class method:

```js

import { randomId } from '@nozbe/watermelondb/utils/common'

randomId.generate = () => Math.random() // number

```
WDYT?

radex:
Yep, agreed with @rkrajewski — that seems simpler

sebastian-schlecht:
Agree with you guys, simple is always good ;) I updated the code, let me know wyt

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ All notable changes to this project will be documented in this file.
 - [TypeScript] 'Cannot use 'in' operator to search for 'initializer'; decorator fix
 
 ### Changes
-
+- [Database] You can now update the random id schema by importing `import { setGenerator } from '@nozbe/watermelondb/utils/common/randomId'` and then calling `setGenerator(newGenenerator)`. This allows WatermelonDB to create specific IDs for example if your backend uses UUIDs.
```
radex: could you please move this to the top (under Unreleased)? Not your fault, just git didn't resolve this correctly after I released newest version :)

radex:
@sebastian-schlecht Thanks for the PR! Looks good to me, just please let's clean up the Changelog :) 

PS. You and your team have contributed a bunch in the discussion threads around 🍉… I'd like to send you a bunch of 🍉 stickers (I didn't do that already, did I?) — If you'd like that, please email me with your address :)

Bessonov:
@rkrajewski https://github.com/Nozbe/WatermelonDB/issues/7#issuecomment-484963108 :)

sebastian-schlecht:
I think we would love to have some stickers! Email incoming :)

sebastian-schlecht:
@radex any ETA on when this is going to be released?:) Planning on using this in our next iteration and I'd love to use the official build :)

radex:
@sebastian-schlecht Probably Monday after the next (next week I'm out of work).

ryan-rokusek:
@sebastian-schlecht I don't understand how to implement setGenerator into my codebase. Would you provide some guidance?

alexandrumic:
@sebastian-schlecht @ryan-rokusek I can't figure it out how to use `setGenerator` into my code, in order to generate random `uuid` to be in sync with the back-end database. I read the changelog but it's a bit unclear. Can you please help me with this? 🤔 

Thank you!

KrisLau:
@ryan-rokusek Where do you call `setGenerator` from? I'm kind of confused about the usage of the method and if it changes the id generated for all tables or just individual tables?

PR #335, opened 2019-04-24
Private fields

rkrajewski:
Partially adds private fields, just to see if they make sense for us.

- No support for decorated private fields - private fields can't be even accessed from decorated fields, such as @lazy.
- Sometimes there is weird problem with eslint, but I wasn't able to figure it out, and had to skip declaring fields as private.
- Some fields are used by helper functions or other internal parts of 🍉 (eg. tests), so I've made extra getters for these, but I'm not sure if this makes more sense, than keeping them as they are.

### DIFF
--- a/src/Collection/index.js
+++ b/src/Collection/index.js
@@ -32,7 +32,7 @@ export default class Collection<Record: Model> {
   #cache: RecordCache<Record>
 
   get _cache(): RecordCache<Record> {
-    invariant(process.env.NODE_ENV === 'test', '_chace can be accessed only in test environment')
+    invariant(process.env.NODE_ENV === 'test', '_cache can be accessed only in test environment')
 
     return this.#cache
   }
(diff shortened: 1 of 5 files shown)


src/Collection/index.js
```diff
@@ -25,24 +25,30 @@ export default class Collection<Record: Model> {
...
-  _cache: RecordCache<Record>
-
   // Emits event every time a record inside Collection changes or is deleted
   // (Use Query API to observe collection changes)
   changes: Subject<CollectionChangeSet<Record>> = new Subject()
 
+  #cache: RecordCache<Record>
+
+  get _cache(): RecordCache<Record> {
+    invariant(process.env.NODE_ENV === 'test', '_chace can be accessed only in test environment')
```
panr:
```suggestion
    invariant(process.env.NODE_ENV === 'test', '_cache can be accessed only in test environment')
```

Typo 😉

radex:
> Some fields are used by helper functions or other internal parts of 🍉 (eg. tests), so I've made extra getters for these, but I'm not sure if this makes more sense, than keeping them as they are.

Hmm :( I think you might be right that it's causing more harm than good :/

PR #342, opened 2019-04-29
Flow fixes

kokusGr:
## List of changes
- Add some Rambdax specific types (piped, mapAsync)
- Fix some types to match Rambdax implementation (map, reduce)
- Add support for generic key in map
- Fix flow issues after changes


kokusGr:
@radex CR pls :)

PR #382, opened 2019-06-19
Add typescript info for sync

jakebloom:
Pulled in the flow definitions and modified them for typescript. This should cover the sync module now!


radex:
Thank you! ❤️

PR #397, opened 2019-06-27
Various improvements

radex:
- [Sync] Performance: `synchronize()` no longer calls your `pushChanges()` function if there are no
    local changes to push. This is meant to save unnecessary network bandwidth. ⚠️ Note that this
    could be a breaking change if you rely on it always being called
- [Sync] When setting new values to fields on a record, the field (and record) will no longer be
    marked as changed if the field's value is the same. This is meant to improve performance and avoid
    unnecessary code in the app. ⚠️ Note that this could be a breaking change if you rely on the old
    behavior. For now you can import `experimentalSetOnlyMarkAsChangedIfDiffers` from
    `@nozbe/watermelondb/Model/index` and call if with `(false)` to bring the old behavior back, but
    this will be removed in the later version -- create a new issue explaining why you need this
- [Sync] Small perf improvements
- [adapters] protecc users from breaking their app by not reading chang…  ……elog


kokusGr:
@radex Uhm, I cannot merge? 🤔 It says only those with *write access* to this repository can merge pull requests 😐

radex:
@kokusGr fixed

kokusGr:
Thx :)

PR #398, opened 2019-06-27
Add support for notLike queries

jakebloom:
Title says it all - added support for `Q.notLike`, in the same vein of `Q.like`.

This was easy enough on sqlite, but on lokijs we had to add a new module, `notLikeToRegexp`, which created a new regex with an inverse match. It looks like this: `/^((?!<expr>).)*$/`

I've added tests to both adapter's `encodeQuery` methods, and mirrored the `like` tests in the `databaseTests`

Here is what the updated doc looks like:
Screen Shot 2019-06-27 at 4 25 11 pm

Feedback much appreciated!


jakebloom:
Hmm okay, I'll fix those tests ☹️

radex:
@jakebloom there are still CI failures :/

docs/Query.md
```diff
@@ -126,7 +126,8 @@ This queries all comments that are **both** verified **and** awesome.
 | `Q.where('likes', Q.between(10, 100))` | `likes >= 10 && likes <= 100` |
 | `Q.where('status', Q.oneOf(['published', 'draft']))` | `status === 'published' \|\| status === 'draft'` |
 | `Q.where('status', Q.notIn(['archived', 'deleted']))` | `status !== 'archived' && status !== 'deleted'` |
-| `Q.where('status', Q.like('%bl_sh%'))` | `/.*bl.sh.*/g` (See note below!) |
+| `Q.where('status', Q.like('%bl_sh%'))` | `/.*bl.sh.*/i` (See note below!) |
+| `Q.where('status', Q.notLike('%bl_sh%'))` | `/^((!?.*bl.sh.*).)*$/i` (Inverse regex match) (See note below!) |
 
 **Note:** It's NOT SAFE to use `Q.like` with user input directly, because special characters like `%` or `_` are not escaped. Always sanitize user input like so:
```
radex:
```suggestion
**Note:** It's NOT SAFE to use `Q.like` and `Q.notLike` with user input directly, because special characters like `%` or `_` are not escaped. Always sanitize user input like so: 
```

native/ios/WatermelonDB.xcodeproj/xcshareddata/xcschemes/WatermelonDB.xcscheme
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
radex: I don't think this should have been committed

radex:
Very cool @jakebloom 🎉 I think there's just a few small tweaks and we can ship this

jakebloom:
Note I had to update the `Nimble` pod in order to build with xcode 10.2

jakebloom:
Looks like the Bridge Tests timed out, they pass on my machine. Closed and reopened the PR in order to rerun the build

jakebloom:
@radex managed to get the tests working - anything else we need for this to get merged?

jakebloom:
@radex anything else we need here?

radex:
Nice! Just negating `like` in loki implementation makes sense to me. I guess perhaps the better long term solution would be to add a negation query term... But that's something to think about another time. Sorry for the delay - was on a conference.

I'm almost ready to merge, but there's one edge case I just thought of: what about `null` behavior? i.e. what happens if we do `LIKE` or `NOT LIKE` against a column which is null?

I think the behavior should be whatever SQLite does, see: https://github.com/Nozbe/WatermelonDB/blob/master/docs/Query.md#null-behavior — would you mind just adding/updating relevant tests in https://github.com/Nozbe/WatermelonDB/blob/e82cee5a52608ec29350969e10bcbda80df33fd9/src/__tests__/databaseTests.js ?

radex:
> Looks like the Bridge Tests timed out, they pass on my machine. Closed and reopened the PR in order to rerun the build

I know, I can't figure out this problem :(

jakebloom:
@radex Good call - I've made those changes and added tests

radex:
Wonderful! Thanks @jakebloom ! I'll release 0.14.0-0 in a moment and if we find no issues in our app, I'll release it officially next week probably

PR #400, opened 2019-06-30
Build docs from markdown and deploy on gh-pages

### LINKED ISSUE #372: Generate static html from markdown docs
Hi,
So I only find there are raw markdown files when learning about this project, I've used mdbook to do such job since reading markdowns with lots of jumping around between different pages on Github is not a very pleasant experience.

you can check it here: https://shadeofgod.github.io/WatermelonDB-docs

Let me know what u think and I'm happy to send a pull request, maybe we can use github pages to host those static files and even set up in CI to automate it.

shadeofgod:
close #372 

@radex 

It will build docs and publish it to GitHub pages after a successful CI run on repository's master branch. But firstly it requires a GitHub `Personal Access Token` with the `public_repo` permissions. You can go to this repository's Travis CI settings page and add an environment variable named `GITHUB_TOKEN` that is marked secure and not shown in the logs.

And please make sure the repo settings on Github have the Github Pages source selected as gh-pages branch after first run. 

I also added a couple extra md files as each chapter's entry for a better catalog(those named with `ch0x-00-xxx.md`).

Another thing worth mentioning is that the tool I am using simply reads ./docs/SUMMARY.md and generates html to ./docs/book, and then travis will deploy the whole ./docs/book directory to gh-pages branch, if the ./docs/SUMMARY.md is referencing any file outside of docs directory, the output that was generated from those files will also be outside of book directroy which then will cause problem for deploying.

So to get things done quickly I simplely copied the whole README.md to ./docs/README.md(not a good idea) and removed referencing CHANGELOG and CONTRUBUTING in SUMMARY.md, let me know if you have any better idea or feel free to edit it by your own after it merged.

also, the tool docs is [here](https://rust-lang-nursery.github.io/mdBook/index.html).


docs/README.md
```diff
@@ -1,42 +1,154 @@
-# WatermelonDB Documentation
+<p align="center">
```
radex: can we maybe add a `cp` to the travis script, so that we don't have to maintain a copy of README?

radex: we could do the same with changelog and contributing maybe? Which I think should stay in the root because some tools auto-recognize such files

.travis.yml
```diff
@@ -71,3 +71,24 @@ matrix:
...
+      cache:
+        - cargo
+      rust:
+        - stable
+      before_script:
+        - (test -x $HOME/.cargo/bin/cargo-install-update || cargo install cargo-update)
+        - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.3" mdbook)
+        - cargo install-update -a
+      script:
+        - mdbook build ./docs
```
radex: one more thought: perhaps it makes sense to move those script items to `script/` folder? So that outside of CI one can easily run this?

radex:
Screen Shot 2019-07-05 at 10 56 07

this looks nice! Thanks @shadeofgod !

docs/README.md
shadeofgod: good idea

.travis.yml
shadeofgod: added a bash script in ./scripts so you can run `yarn run docs`

shadeofgod:
@radex fixed

radex:
Thanks a lot! I'll check this out on Friday

radex:
@shadeofgod I took a slightly different approach in the end - instead of auto-deploying on travis onto a separate branch, I keep the compiled book on master/docs, and I update it in the publish script -- this is so that the official docs don't show the "too new" features before they're released.

Thank you!

shadeofgod:
@radex sure, if you want to manually depoy in a single command you can still put those compiled files in gh-pages branch by using a package named [gh-pages](https://www.npmjs.com/package/gh-pages) so the master branch would be a bit cleaner.

radex:
@shadeofgod oh nice! Thanks!

PR #447, opened 2019-08-10
fix(Schema): handle invalid table schema argument in appSchema

### LINKED ISSUE #446: `Cannot read property 'name' of undefined` when passing an invalid object to tables array
Hi, there's an unhandled error when creating an `appSchema` object, when you mistakenly pass an invalid type to the `tables` array, you get

```
Cannot read property 'name' of undefined
```

which isn't telling you enough about the issue.

### How to reproduce

```js
import { appSchema } from '@nozbe/watermelondb'

let invalidSchema = undefined

let rootSchema = appSchema({
  version: 1,
  tables: [invalidSchema],
})
```

krszwsk:
Fixes #446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- [Schema] Handle invalid table schema argument in appSchema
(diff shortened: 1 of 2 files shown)


radex:
@krszwsk Thank you for improving DX for other library users as you stumble upon issues!

Issue #453, opened 2019-08-17
[bug?] global is not defined


roytan883:
Hi, i try to use WatermelonDB in electron on windows. But I got this error:
```
Uncaught ReferenceError: global is not defined
    at index.js:15
    at Object../node_modules/@nozbe/watermelondb/utils/common/devMeasureTime/index.js (index.js:22)
```
after I modify this file:
> node_modules/@nozbe/watermelondb/utils/common/devMeasureTime/index.js

```
//if (global && global.nativePerformanceNow) {
if ('undefined' !== typeof global && global.nativePerformanceNow) {
```

**it works**.

radex:
@roytan883 great, send a pull request!

PR #458, opened 2019-08-20
fix can't run in electron

roytan883:
to avoid `Uncaught ReferenceError: global is not defined` in electron runtime.


roytan883:
#453

radex:
@roytan883 can you add a note about this to CHANGELOG?

radex:
Thanks @roytan883 ! Please do send more pull requests if you come across any other issues :)

PR #473, opened 2019-08-28
[rambdax] Use includes instead of contains

### LINKED ISSUE #471: TypeError: (0 , _rambdax.contains) is not a function
Hello, so I am receiving the following error `TypeError: (0 , _rambdax.contains) is not a function` at `@nozbe/watermelondb/Schema/index.js:48` where the verification ` (0, _invariant.default)((0, _rambdax.contains)(column.type, ['string', 'boolean', 'number']), "Invalid type ".concat(column.type, " for column ").concat(column.name, " (valid: string, boolean, number)"));` is done, after further investigation I saw that the Rambdax module received a minor bump and in the previous commits it replaced the `contains` function with `includes` please check here the commit [https://github.com/selfrefactor/rambdax/commit/789dad193d83fcc5aec5f95cd3b06f2e4eb35ba4](url) can the code be updated? I think the only change is to replace `contains` with `includes`, not sure since I didn't check for any other breaking change.

rkrajewski:
- Replaces `contains` with `includes` due to `contains` deprecation https://github.com/selfrefactor/rambda/commit/1dc1368f81e9f398664c9d95c2efbc48b5cdff9b#diff-04c6e90faac2675aa89e2176d2eec7d8R2209
- Resolves #471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 - [Schema] Handle invalid table schema argument in appSchema
 - [withObservables] Added TypeScript support ([changelog](https://github.com/Nozbe/withObservables/blob/master/CHANGELOG.md))
 - [Electron] avoid `Uncaught ReferenceError: global is not defined` in electron runtime ([#453](https://github.com/Nozbe/WatermelonDB/issues/453))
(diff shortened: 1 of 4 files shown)


radex:
thanks @rkrajewski !

PR #505, opened 2019-09-27
LokiJS options

radex:
### New featuers

- [LokiJS] Introduces new `new LokiJSAdapter({ ..., experimentalUseIncrementalIndexedDB: true })` option.
  When enabled, database will be saved to browser's IndexedDB using a new adapter that only saves the
  changed records, instead of the entire database.

  **This works around a serious bug in Safari 13** (https://bugs.webkit.org/show_bug.cgi?id=202137) that causes large
  databases to quickly balloon to gigabytes of temporary trash

  This also improves performance of incremental saves, although initial page load or very, very large saves
  might be slightly slower.

  This is intended to become the new default option, but it's not backwards compatible (if enabled, old database
  will be lost). **You're welcome to contribute an automatic migration code.**

  Note that this option is still experimental, and might change in breaking ways at any time.

- [LokiJS] Introduces new `new LokiJSAdapter({ ..., useWebWorker: false })` option. Before, web workers
  were always used with `LokiJSAdapter`. Although web workers may have some performance benefits, disabling them
  may lead to lower memory consumption, lower latency, and easier debugging. YMMV.

### Improvements

- [LokiJS] Persistence adapter will now be automatically selected based on availability. By default,
  IndexedDB is used. But now, if unavailable (e.g. in private mode), ephemeral memory adapter will be used.


radex:
⚠️ please don't merge yet - waiting confirmation on NT front that there are no regressions

but please CR - if anything, only minor changes will come to this PR

radex:
@kokusGr Can you review this week?

src/adapters/lokijs/worker/lokiExtensions.js
```diff
@@ -1,39 +1,79 @@
...
+    if (typeof indexedDB === 'undefined') {
+      resolve(false)
+    }
 
-  return new Loki(name, {
-    adapter: peristenceAdapter || newAdapter,
-    autosave: true,
-    autosaveInterval: 250, // TODO: Remove this and force database save when we have transactions
-    env: 'BROWSER', // TODO: ?
-    verbose: true, // TODO: remove later
```
kokusGr: Are these comments no longer valid? I don't see them in new version :)

src/adapters/lokijs/worker/lokiExtensions.js
```diff
@@ -1,39 +1,79 @@
...
 import Loki, { LokiMemoryAdapter } from 'lokijs'
-import LokiIndexedAdapter from 'lokijs/src/loki-indexed-adapter'
 
-export function newLoki(name: ?string, peristenceAdapter?: LokiMemoryAdapter): Loki {
-  const newAdapter =
-    process.env.NODE_ENV === 'test' ? new LokiMemoryAdapter() : new LokiIndexedAdapter(name)
+const isIDBAvailable = () => {
+  return new Promise(resolve => {
+    // $FlowFixMe
+    if (typeof indexedDB === 'undefined') {
```
kokusGr: What's the problem with Flow?

src/adapters/lokijs/worker/lokiExtensions.js
radex: verbose doesn't really do much of anything - IIRC it only spits out errors to the console, and env is not necessary / doesn't do anything if you pass adapter explicitly

src/adapters/lokijs/worker/lokiExtensions.js
radex: it doesn't know that `indexedDB` is defined, and I can't do `window.indexedDB` because this code can run either in normal context (has window) or in a web worker (doesn't have window)

src/adapters/lokijs/worker/lokiExtensions.js
kokusGr: Cool, just wanted to make sure these weren't omitted by mistake :)

src/adapters/lokijs/worker/lokiExtensions.js
kokusGr: Ok, we can leave it as is :)

PR #513, opened 2019-10-06
Android dependecies update

rozPierog:
Updated Kotlin, Gradle and native dependecies in Library, Test app, and Example App. 
Also moved Example app to AndroidX - as it is standard right now.


rozPierog:
Might have broken example app with those updates. Gonna fix it up later 😶

PR #518, opened 2019-10-11
upate web demo to use withDatabase hoc

kriswep:
As suggested in #141 this PR updates the web example to use `withDatabase` and `DatabaseContext`. This reduces the `database` prop drilling.

Everything else, especially the native demo, is untouched.


radex:
thanks @kriswep

Issue #487, opened 2019-09-12
Model instance dependent json sanitisation


davidkell:
We have a use case with the following parameters:
- There is a single model `MyEntity`
- `MyEntity` has a `kind` field (an Enum) and a `content` field (JSON)
- The JSON schema for the `content` field depends on the `kind`

We want to validate and sanitise the JSON field using the sanitise content field. Is it possible to provide information on the `kind` to the sanitise content function for the `content` JSON field? (We couldn't find a way.)

A more concrete example:

```
export default class MyEntity extends Model {
  static table = 'myentity'

  @field('kind') kind

  @json('content', (content: object) => {
    // Here we want the validation function to depend on the kind.
    // Unfortunately this is not defined.
    validate(this.kind, content)
    return content
  })

}
```

(Another solution would be to have separate model for each distinct kind. Unfortunately, this would not work as we need to dynamically generate new kinds.)

rkrajewski:
@davidkell Hi, indeed there is no way to do it, maybe only if data provided to sanitizer would contain a `kind`, but probably it's not what you look for.

Feel free to contribute - at first look, only what you need is to call sanitizer with `this` context.

here https://github.com/Nozbe/WatermelonDB/blob/master/src/decorators/json/index.js#L42
and here https://github.com/Nozbe/WatermelonDB/blob/master/src/decorators/json/index.js#L45

```js
sanitizer.call(this, json)
```

but I haven't tested this solution.

davidkell:
Thanks, will have a look at this.

PR #526, opened 2019-10-17
Reference the model within JSON sanitizer

davidkell:
Adds a reference to the model within the JSON sanitizer via `this`, as suggested here #487

### DIFF
--- a/src/decorators/json/index.js
+++ b/src/decorators/json/index.js
@@ -39,10 +39,10 @@ export const jsonDecorator = makeDecorator(
         const rawValue = this.asModel._getRaw(rawFieldName)
         const parsedValue = parseJSON(rawValue)
 
-        return sanitizer(parsedValue)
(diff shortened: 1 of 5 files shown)


src/decorators/json/index.js
```diff
@@ -39,10 +39,10 @@ export const jsonDecorator = makeDecorator(
         const rawValue = this.asModel._getRaw(rawFieldName)
         const parsedValue = parseJSON(rawValue)
 
-        return sanitizer(parsedValue)
+        return sanitizer.call(this, parsedValue)
```
radex: I'm not comfortable with passing the Modal as `this`... Feels too magical. Why not just add second argument?

davidkell: Yes, agreed! Just pushed the changes.

radex:
Nice, thanks @davidkell !

PR #529, opened 2019-10-18
fix: typescript error

johannesjo:
This way using this via typescript will at least not throw an error.

No clue what this is about:
```  
// TODO: fix
> = ReplaceReturn<Args, Descriptor | RawDecorator, Factory>
```

I help if I can.

### DIFF
--- a/src/utils/common/makeDecorator/index.d.ts
+++ b/src/utils/common/makeDecorator/index.d.ts
@@ -1,5 +1,5 @@
 declare module '@nozbe/watermelondb/utils/common/makeDecorator' {
-  // import { ReplaceReturn } from '@nozbe/watermelondb/utils/common'
+  import { ReplaceReturn } from '@nozbe/watermelondb/utils/common'
 
   export type Descriptor = Object
   export type RawDecorator = (target: Object, key: string, descriptor: Descriptor) => Descriptor


tomdaniel0x01:
> Hello I checked the latest release and it seems now that this fix causes all decorators `@field @date` etc. to throw the following error:
> 
> ```
> Unable to resolve signature of property decorator when called as an expression.
>   This expression is not callable.
>     Not all constituents of type 'Object | RawDecorator' are callable.
>       Type 'Object' has no call signatures.
> 
> 82     @field(VehicleColumns.milesPerDay) milesPerDay?: string;
> ```
> 
> It seems that this issue was not caught until now due to the commented import and by defauld typescriopt not really validating the decorators. Not that they are validated I receive the following error in the entire application. I managet to resolve it by changing the return type of the Descriptor to any:
> 
> ```
> export type Descriptor = any
> ```
> 
> instead of:
> 
> ```
> export type Descriptor = Object
> ```
> 
> not sure what the correct solution would be here. I also tried the compiler option solutions here but they were already set:
> [StackOverflow](https://stackoverflow.com/questions/37694322/typescript-ts1241-unable-to-resolve-signature-of-method-decorator-when-called-a)
> 
> Also after the fix the `@readonly` decorator gives the error:
> 
> ```
> error TS1240: Unable to resolve signature of property decorator when called as an expression.
> 
> 176     @readonly @date(GeneralTableColumns.createdAt) createdAt!: Date;
> ```
> 
> If I change the readonly decorator call to `@readonly()` it doesn't complain but I don't know if it is correct.
> 
> @radex should this be reported as a separate issue or is it ok here since this fix revealed other TypeScript issues?

I am also receiving this error when using the `@field` decorator.

PR #534, opened 2019-10-21
Fix observeWithColumns glitches

radex:
- [x] fixes fieldObserver / observeWithColumns race condition which caused weird glitches - e.g. in NT when removing a star from task in Priority, task would first go to the bottom, only then disappear


src/observation/observeQueryWithColumns/test.js
```diff
@@ -0,0 +1,163 @@
+import { mockDatabase } from '../../__tests__/testModels'
+import * as Q from '../../QueryDescription'
+import fieldObserver from './index'
```
kokusGr: You forgot to change the name in tests :)

radex:
done

PR #547, opened 2019-10-30
Perf improvements

radex:
- Add function names for Rx observers and other things triggered from microtask queue for easier debugging (hard to understand what's going on in profiles or stack traces when it's just RxJS bullshit and `(anonymous)`)
- Some de-rx-ification of the code. the code looks worse :( … but seriously, rx stack traces are so enormous, it's hard to work with profiles
- disable most logging, because it's useless & slow
- schedule Loki app→executor communication in `useWebWorker: false` mode as Microtasks, not Macrotasks. This way, we can make UI synchronous-ish -- control won't be returned to browser to paint before the response arrives, so we can avoid glitches. In the future, it would be nice to take good advantage of multithreading/concurrency with web workers -- that was a selling feature of 🍉 after all -- but as things stand today, Loki queries are SUPER fast, and rendering is the bottleneck. So doing things synchronously(ish) and on one thread is faster.

### DIFF
--- a/src/adapters/lokijs/worker/workerMock.js
+++ b/src/adapters/lokijs/worker/workerMock.js
@@ -32,7 +32,9 @@ export default class LokiWorkerMock {
     this._workerContext = {
       postMessage: data => {
         const clonedData = clone(data)
-        setImmediate(() => {
+
+        // Schedule on the microtask queue to be able to achieve synchronous, glitch-free rendering
+        Promise.resolve().then(() => {
           this.onmessage({ data: clonedData })
         })
       },
@@ -58,7 +60,7 @@ export default class LokiWorkerMock {
       clonedData = clone(data)
     }
 
-    setImmediate(() => {
+    Promise.resolve().then(() => {
       // $FlowFixMe
       this._workerContext.onmessage({ data: clonedData })
     })
--- a/src/observation/simpleObserver/index.js
+++ b/src/observation/simpleObserver/index.js
@@ -66,7 +66,9 @@ function observeChanges<Record: Model>(
       emitCopy()
 
       // Observe changes to the collection
-      return query.collection.changes.subscribe(changeSet => {
+      return query.collection.changes.subscribe(function observeQueryCollectionChanged(
+        changeSet,
+      ): void {
         const shouldEmit = processChangeSet(changeSet, matcher, matchingRecords)
         if (shouldEmit || alwaysEmit) {
           emitCopy()
--- a/src/observation/simpleObserver/index.js
+++ b/src/observation/simpleObserver/index.js
@@ -1,13 +1,8 @@
 // @flow
 
 import { Observable } from 'rxjs/Observable'
-import { defer } from 'rxjs/observable/defer'
-import { switchMap } from 'rxjs/operators'
 
-import doOnDispose from '../../utils/rx/doOnDispose'
-import doOnSubscribe from '../../utils/rx/doOnSubscribe'
-
-import logger from '../../utils/common/logger'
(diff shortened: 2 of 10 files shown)


src/adapters/lokijs/worker/workerMock.js
```diff
@@ -32,7 +32,9 @@ export default class LokiWorkerMock {
     this._workerContext = {
       postMessage: data => {
         const clonedData = clone(data)
-        setImmediate(() => {
+
+        // Schedule on the microtask queue to be able to achieve synchronous, glitch-free rendering
+        Promise.resolve().then(() => {
```
kokusGr: Hmm wouldn't `requestAnimationFrame` be better? It's guaranteed to run before the next frame while microtasks kinda are more unpredictable running only when stack is empty.

src/observation/simpleObserver/index.js
```diff
@@ -52,38 +47,48 @@ export function processChangeSet<Record: Model>(
...
 ): Observable<Record[]> {
-  return defer(() => query.collection.fetchQuery(query)).pipe(
-    switchMap(observeChanges(query, alwaysEmit)),
-    doOnSubscribe(() => logger.log(`Subscribed to changes in a ${query.table} query`)),
-    doOnDispose(() => logger.log(`Unsubscribed from changes in a ${query.table} query`)),
-  )
+  // Note: it would be cleaner to do defer->switchMap, but that makes profiles really hard to read
+  // hence the mutability
+  return Observable.create(observer => {
+    // logger.log(`Subscribing to changes in a ${query.table} query`)
```
kokusGr: Is this line still needed?

src/observation/simpleObserver/index.js
```diff
@@ -1,13 +1,8 @@
...
 
 import { Observable } from 'rxjs/Observable'
-import { defer } from 'rxjs/observable/defer'
-import { switchMap } from 'rxjs/operators'
 
-import doOnDispose from '../../utils/rx/doOnDispose'
-import doOnSubscribe from '../../utils/rx/doOnSubscribe'
-
-import logger from '../../utils/common/logger'
+// import logger from '../../utils/common/logger'
```
kokusGr: Is this line still needed?

src/adapters/lokijs/worker/workerMock.js
radex: Hmm, 

> microtasks kinda are more unpredictable running only when stack is empty.

I don't know anything about that, but my understanding is `rAF` is also unpredictable because spec is not clear about when this should be executed, and if main thread is busy for a long time, it won't be executed immediately, given that its purpose is to schedule… stuff before next animation frame ;) https://stackoverflow.com/a/43050651

I will continue doing research on this and get back on you

src/observation/simpleObserver/index.js
radex: i guess not don

src/observation/simpleObserver/index.js
radex: i guess not ✅

radex:
@kokusGr Merge?

src/adapters/lokijs/worker/workerMock.js
kokusGr: Yeah I also haven't tested this and my comment was based on things I read/heard. AFAIK `rAF` should run between macrotasks so if 1 task takes a lot of time a frame probably will be skipped. However if thread is blocked by 1 task, microtasks also probably won't run 🤔 But unfortunately this can differ between browsers 🙁

kokusGr:
Yup, merging thx :)

PR #563, opened 2019-11-12
(iOS) Experimental pre-JSI synchronous SQLiteAdapter

radex:
- [SQLiteAdapter][iOS] Add new `synchronous` option to adapter: `new SQLiteAdapter({ ..., synchronous: true })`.
   When enabled, database operations will block JavaScript thread. Adapter actions will resolve in the
   next microtask, which simplifies building flicker-free interfaces


src/adapters/sqlite/index.js
```diff
@@ -64,13 +81,79 @@ type NativeBridgeType = {
...
+  'setLocal',
+  'removeLocal',
+]
+
+type NativeBridgeType = {
+  // Async methods
+  ...NativeDispatcher,
+
+  // Synchronous methods
+  initializeSync?: (ConnectionTag, string, SchemaVersion) => SyncReturn<InitializeStatus>,
```
rozPierog: I'm not a fan of `*Sync` naming scheme, it seems confusing especialy when we are talking about database that can `Synchronize` with server. Maybe something like `Synch` 🤷🏻‍♂️

src/adapters/sqlite/integrationTest.js
```diff
@@ -1,18 +1,47 @@
...
-        const adapter = newAdapter()
+        const adapter = new SQLiteAdapter({ schema: testSchema, synchronous: isSynchronous })
+
+        if (isSynchronous) {
+          if (Platform.OS === 'ios') {
+            invariant(adapter._synchronous === true, 'this should be synchronous')
+          } else {
+            invariant(
+              adapter._synchronous === false,
+              'this should be asynchronous - android does not support synchronous adaoter',
```
rozPierog:
```suggestion
              'this should be asynchronous - android does not support synchronous adapter',
```

rozPierog: Also, we should avoid creating such discrepancy between Android and iOS. We all know how frustrating is when you are trying to implement something but for some reason it works only on one platform

radex: True, OTOH: a) we agreed on Slack that it's not worth pursuing this internally on Android with current performance levels, b) it's an option you have to know about to use, c) there's a graceful fallback

src/adapters/sqlite/index.js
radex: riiiight. I'll go with `*Synchronous`

radex:
fixed - let's wait for Travis to confirm i didn't mess anything up

rozPierog:
Thanks 🎉

Issue #564, opened 2019-11-12
0.15 unsafeResetDatabase not finishing?


ospfranco:
I'm testing the 0.15 update and some of our tests got broken:

![Screenshot 2019-11-12 at 14 47 42](https://user-images.githubusercontent.com/1634213/68677057-a2388f80-055b-11ea-8263-1cf433d28604.png)

Very simple test that just mounts a sample lokiJS database and tries to wipe it.

Any suggestions?

radex:
are you using new LokiJSAdapter modes?

ospfranco:
```
import uuidv4 from 'uuid/v4';
import { Database } from '@nozbe/watermelondb';
import LokiJSAdapter from '@nozbe/watermelondb/adapters/lokijs';

const setupTestAdapter = (schema: any, dbName?: string) => {
  // First, create the adapter to the underlying database:
  const adapter = new LokiJSAdapter({
    dbName: dbName ? dbName : uuidv4(),
    schema,
  });

  return adapter;
};

const setupTestDb = (adapter: any, modelClasses: any) => {
  // Then, make a Watermelon database from it!
  const database = new Database({
    adapter,
    modelClasses,
    actionsEnabled: true,
  });

  return database;
};

export { setupTestAdapter, setupTestDb };
```

This is the set up code

ospfranco:
I added some logging into the watermelon code, here is what I found:

```
console.warn node_modules/@nozbe/watermelondb/Database/index.js:203
    clear caches DONE
// All the steps pass except the unsafe database reset call to the worker thread

  console.warn node_modules/@nozbe/watermelondb/adapters/lokijs/WorkerBridge.js:50
    sending UNSAFE_RESET_DATABASE []
// Added an extra unsafe reset database call to see what happens

  console.warn node_modules/@nozbe/watermelondb/adapters/lokijs/WorkerBridge.js:50
 sending UNSAFE_RESET_DATABASE []
// The original unsafe reset database call message is sent

  console.log node_modules/@nozbe/watermelondb/utils/common/logger/index.js:21
    [DB][Worker] Initializing IndexedDB  console.log node_modules/@nozbe/watermelondb/utils/common/logger/index.js:21
    [DB][Worker] Database loaded

// The database just then starts loading??

  console.log node_modules/@nozbe/watermelondb/utils/common/logger/index.js:21
    [DB][Worker] Empty database, setting up

[...]

PR #565, opened 2019-11-12
Fix delete database code, loki.close does not take a callback

ospfranco:
Takes care of #564

### DIFF
--- a/src/adapters/lokijs/worker/lokiExtensions.js
+++ b/src/adapters/lokijs/worker/lokiExtensions.js
@@ -31,7 +31,7 @@ async function getLokiAdapter(
   name: ?string,
   adapter: ?LokiMemoryAdapter,
   useIncrementalIDB: boolean,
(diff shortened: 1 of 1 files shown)


radex:
wonderful! thanks @ospfranco for doing the work!

See, there's still an issue here:

https://github.com/techfort/LokiJS/blob/master/src/lokijs.js#L1807-L1818 - the callback should be called. And it's important to use it, because if database is dirty, we want it to save before DB is deleted. Otherwise, since it may be asynchronous, it maybe save AFTER it's deleted. which is not good.

perhaps there's something wrong here: https://github.com/techfort/LokiJS/blob/master/src/lokijs.js#L1987

ospfranco:
Yeah, I understood it is important to prevent race conditions, I probably won't have the time to look into lokiJS, at least with this PR unsafeResetDatabase starts working again.

I monkey patched our local version, but I guess this is an issue for the v0.15 release.

Issue #572, opened 2019-11-21
Using same server ID in local database


euharrison:
Hi, thanks for this wonderful lib! 👏 

I already read a lot of issues and PRs but I'm not 100% sure if I can use the same `server id` in local database, my main points are:

- Is there a conflict using a non-global id?
- Is the cache working well for non-global id?

Probably this PR https://github.com/Nozbe/WatermelonDB/pull/175 enable it, but I just need to be sure that it's really working

![Screen Shot 2019-11-21 at 16 05 25](https://user-images.githubusercontent.com/557598/69368753-7911b080-0c79-11ea-95f6-78cacc60c9e4.png)
![Screen Shot 2019-11-21 at 16 05 31](https://user-images.githubusercontent.com/557598/69368754-79aa4700-0c79-11ea-8bde-247cfcf6204f.png)

radex:
> Probably this PR #175 enable it, but I just need to be sure that it's really working

it should - is it not working for you?

euharrison:
Yes, it's!

Thanks for reply! I just need to be 100% sure

Do you suggest to add the tip of use `record._raw.id` instead of `record.id` into the docs?

radex:
> Do you suggest to add the tip of use record._raw.id instead of record.id into the docs?

sorry, i'm a little confused what you're referring to exactly

euharrison:
I opened the PR it's easier to explain :)
https://github.com/Nozbe/WatermelonDB/pull/574

Please, feel free to close it if too specific or suggest something else.

PR #574, opened 2019-11-25
Add post._raw.id docs

euharrison:
Ref: https://github.com/Nozbe/WatermelonDB/issues/572

Add docs of how to override `record.id`


radex:
thanks!

PR #577, opened 2019-11-28
de-Rx-ify

radex:
- [Model] Added experimental `model.experimentalSubscribe((isDeleted) => { ... })` method as a vanilla JS alternative to Rx based `model.observe()`. Unlike the latter, it does not notify the subscriber immediately upon subscription.
- [Collection] Added internal `collection.experimentalSubscribe((changeSet) => { ... })` method as a vanilla JS alternative to Rx based `collection.changes` (you probably shouldn't be using this API anyway)
- [Database] Added experimental `database.experimentalSubscribe(['table1', 'table2'], () => { ... })` method as a vanilla JS alternative to Rx-based `database.withChangesForTables()`. Unlike the latter, `experimentalSubscribe` only emits notifies the subscriber once after a batch that makes a change in multiple collections subscribed to. It also doesn't notify the subscriber immediately upon subscription, and doesn't send details about the changes, only a signal.
- Added `experimentalDisableObserveCountThrottling()` to `@nozbe/watermelondb/observation/observeCount` that globally disables count observation throttling. We think that throttling on WatermelonDB level is not a good feature and will be removed in a future release - and will be better implemented on app level if necessary

### Changes

- [LokiJS] [Performance] Improved worker queue implementation for performance
- [observation] Refactored observer implementations for performance


src/Collection/index.js
```diff
@@ -153,6 +156,17 @@ export default class Collection<Record: Model> {
...
   }
 
+  _subscribers: Array<(CollectionChangeSet<Record>) => void> = []
+
+  experimentalSubscribe(subscriber: (CollectionChangeSet<Record>) => void): () => void {
+    this._subscribers.push(subscriber)
+
+    return () => {
+      const idx = this._subscribers.indexOf(subscriber)
+      this._subscribers.splice(idx, 1)
```
kokusGr: Shouldn't we check if `idx !== -1`? You can call this function many times

src/Model/test.js
```diff
@@ -716,6 +716,51 @@ describe('Model observation', () => {
...
+    expect(observer3).toHaveBeenCalledTimes(2)
+    expect(observer3).toHaveBeenLastCalledWith(false)
+
+    await task.markAsDeleted()
+
+    expect(observer3).toHaveBeenCalledTimes(3)
+    expect(observer3).toHaveBeenLastCalledWith(true)
+
+    unsubscribe3()
+    unsubscribe3()
```
kokusGr: We could also test if calling `unsubscribe3` won't delete other subscribers sth like this:
```
const observer1 = jest.fn()
const unsubscribe1 = task.experimentalSubscribe(observer1)
const observer2 = jest.fn()
const unsubscribe2 = task.experimentalSubscribe(observer2)
const observer3 = jest.fn()
const unsubscribe3 = task.experimentalSubscribe(observer3)

unsubscribe3()
unsubscribe3()
unsubscribe3()

expect(task._subcribers.length == 2)
expect(task._subscribers).toEqual([observer1, observer2])
```

src/Database/index.js
```diff
@@ -126,6 +134,23 @@ export default class Database {
     return merge$(...changesSignals).pipe(startWith(null))
   }
 
+  _subscribers: Array<[TableName<any>[], () => void]> = []
+
+  // Notifies `subscriber` on change in any of passed tables (only a signal, no change set)
```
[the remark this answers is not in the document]
radex: Because it was more work for me, and for the computers, and currently we have no use case where we actually need to subscribe to actual changes for multiple tables, just the signal that we need to re-query the database. We can always add it later without making backwards-incompatible changes.

src/Collection/index.js
radex: shit, you're right here. I thought I checked that splicing with `-1` as first arg is a no-op, but… it's not. will fix.

src/Database/index.js
```diff
@@ -126,6 +134,23 @@ export default class Database {
...
+  experimentalSubscribe(tables: TableName<any>[], subscriber: () => void): () => void {
+    if (!tables.length) {
+      return noop
```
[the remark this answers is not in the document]
radex: if you don't pass an inline array to `experimentalSubscribe`, but just pass a variable from some other code, you could get there… not sure why, but i also don't see that it would be harmful

src/Collection/index.js
radex: fixed for all subscribers

src/Model/test.js
radex: I wrote tests that are based on behavior, not checking private properties

kokusGr: That's true, you are right. However it could be written like that:
```
const observer1 = jest.fn()
const unsubscribe1 = task.experimentalSubscribe(observer1)
const observer2 = jest.fn()
const unsubscribe2 = task.experimentalSubscribe(observer2)
const observer3 = jest.fn()
const unsubscribe3 = task.experimentalSubscribe(observer3)

unsubscribe3()
unsubscribe3()
unsubscribe3()

await task.update()

expect(observer1).toHaveBeenCalledTimes(1)
expect(observer2).toHaveBeenCalledTimes(1)
expect(observer3).toHaveBeenCalledTimes(0)
```
It's testing a behavior and checks if multiple unsubscribes don't mess other subscriptions.

PR #588, opened 2019-12-13
Add get method in database model

rahulrana95:


src/Database/index.js
```diff
@@ -53,6 +53,10 @@ export default class Database {
     this._actionsEnabled = actionsEnabled
   }
 
+  get(query = ''): Array<T> {
```
radex: this doesn’t type-check. I want to have this API, but the method needs to pass Flow checks, needs to have a rudimentary test (Database/test.js), TypeScript support would be nice (Database/index.t.ts), and a single-line comment explaining what this is

rahulrana95: Ok will do proper Flow checks, also will write test too.

radex: @rahulrana95 are you planning to finish this PR? :)

PR #592, opened 2019-12-16
Bisynchronous adapter

radex:
So I implemented the Promise → callback switch in Adapters API to allow for bisynchronicity.

Pretty quickly I realized though that there's a ton of direct adapter calls, especially in test code, and it's going to be a huge pain in the ass to rewrite that all from async/await to dealing with callbacks.

The solution I came up with is I implemented `DatabaseAdapterCompat` which implements the OLD, unchanged Adapter interface. So using `database.adapter.xxxx(....)` requires no changes in code. When you want to use callback API for performance, you can go down to `database.adapter.underlyingAdapter.xxx(...)`.

It's not perfect… We add yet another layer of abstraction, and 🍉's complexity grows. There's also some side effects in test code where I need to drop down to underlyingAdapter or be careful about adapter handling when testCloning. But it makes the transition much easier overall, especially in light of the idea that while this is very profitable and worth pursuing now, we might decide in 6-12 months to revert many of those changes if React Suspense + async methods proves to be as performant.

I added some very lightweight `Result` monad types & operations.

### DIFF
--- a/src/adapters/lokijs/worker/lokiWorker.js
+++ b/src/adapters/lokijs/worker/lokiWorker.js
@@ -52,7 +52,7 @@ export default class LokiWorker {
   executeNext(): void {
     const action = this.queue[0]
     const onActionDone = (response: WorkerResponse): void => {
-      invariant(this._actionsExecuting === 1, 'worker queue should have 1 item')
+      invariant(this._actionsExecuting === 1, 'worker should be executing 1 action')
       this._actionsExecuting = 0
       this.queue.shift()
 
@@ -63,7 +63,7 @@ export default class LokiWorker {
       }
     }
 
-    invariant(this._actionsExecuting === 0, 'worker queue should be empty') // sanity check
+    invariant(this._actionsExecuting === 0, 'worker should not have ongoing actions') // sanity check
     this.processAction(action, onActionDone)
   }
 
--- a/src/adapters/lokijs/worker/lokiWorker.js
+++ b/src/adapters/lokijs/worker/lokiWorker.js
@@ -51,28 +51,54 @@ export default class LokiWorker {
 
   executeNext(): void {
     const action = this.queue[0]
-    const onActionDone = (response: WorkerResponse): void => {
-      invariant(this._actionsExecuting === 1, 'worker should be executing 1 action')
-      this._actionsExecuting = 0
-      this.queue.shift()
-
-      this.workerContext.postMessage(response)
-
-      if (this.queue.length) {
-        this.executeNext()
-      }
-    }
-
     invariant(this._actionsExecuting === 0, 'worker should not have ongoing actions') // sanity check
-    this.processAction(action, onActionDone)
+    this.processAction(action)
   }
 
-  async processAction(action: WorkerAction, callback: WorkerResponse => void): Promise<void> {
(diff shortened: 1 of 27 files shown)


src/observation/subscribeToQueryWithColumns/index.js
```diff
@@ -60,7 +60,7 @@ export default function subscribeToQueryWithColumns<Record: Model>(
   // on the other, it would be good to have source provided as Observable, not Query
   // so that we can reuse cached responses -- but they don't have compatible format
   const [subscribeToSource, asyncSource] = query.hasJoins
-    ? [observer => subscribeToQueryReloading(query, observer, true), true]
+    ? [observer => subscribeToQueryReloading(query, observer, true), false]
```
radex: this is now slightly broken (performance/flicker optimization from v0.15 is reverted if you DON'T use synchronous mode). Not sure if this should be a blocker. Your call.

src/adapters/lokijs/worker/lokiWorker.js
```diff
@@ -51,29 +49,55 @@ export default class LokiWorker {
...
       invariant(type in actions, `Unknown worker action ${type}`)
 
+      if (type === actions.SETUP || type === actions.UNSAFE_RESET_DATABASE) {
+        this.processActionAsync(action)
+      } else {
+        // run action
+        invariant(this.executor, `Cannot run actions because executor is not set up`)
+
+        const runExecutorAction = executorMethods[type].bind(this.executor)
+        const response = runExecutorAction(...payload)
```
kokusGr: Hmm it's very similar to else block in `processActionAsync`. The only difference is `await` keyword. Maybe move this to helper method to avoid code duplication?

src/observation/subscribeToQueryWithColumns/index.js
kokusGr: Hmm I think this is ok. Updating library is not mandatory so if someone feels strongly about this they can use v0.15 until this issue is resolved.

src/observation/subscribeToSimpleQuery/index.js
```diff
@@ -60,28 +60,33 @@ export default function subscribeToSimpleQuery<Record: Model>(
...
-    .then(function observeQueryInitialEmission(initialRecords): void {
-      if (unsubscribed) {
-        return
-      }
+  query.collection._fetchQuery(query, function observeQueryInitialEmission(result): void {
+    if (unsubscribed) {
+      return
+    }
+
+    if (!result.value) {
```
kokusGr: Is there a specific reason why are we checking if value is undefined instead of checking if `result.error` is defined?

radex: no reason. checking for `result.error` is more correct, although in this case value will never be falsy. will fix

src/adapters/lokijs/worker/lokiWorker.js
radex: ✅

radex:
I fixed the issues you've raised in next PR, since they're all refactoring, not bugs: https://github.com/Nozbe/WatermelonDB/pull/593/files

radex:
merge?

kokusGr:
Sure, thanks :)

Issue #51, opened 2018-09-23
Is watermelon safe to store sensitive data?


ikhsanalatsary:
I just curious about it.

rozPierog:
Android approach to this is not great, to say at least. All database files are kept in apps `/data/` folder from where they are free to grab for any rooted user. (non-root apps _should_ not be able to grab that file) That file can be read by any sql file reader. So if you want better than native security I would suggest some kind of hashing of data before inserting it in database or even better - storing sensitive data on your own server and not on a device.

radex:
We should make some research into SQLite encryption then @rozPierog — I assume there's some Keychain-like API on Android for storing sensitive keys (like a database encryption key), right? And I know that SQLite has encryption capabilities on top of it (but we'd have to bundle our own version of sqlite into the app in that case)

rozPierog:
@radex there is [Keystore](https://developer.android.com/training/articles/keystore), I've just skimmed thrugh it and it seems like good equvalent of iOS Keychain. Bundling our own SQLite can be expensive on apk size and performance, if we are going to do that that should be an option. Not everyone stores sensitive data on device.

radex:
@rozPierog OK — can you add this to 🍉 backlog in Nozbe? We should probably research this for our purposes.

radex:
@ikhsanalatsary Does this answer your questions? If so, please close the issue :)

ikhsanalatsary:
@radex Yes, I hope it will implement it in this lib. thank you so much for the answer

waqas19921:
@rozPierog @radex can we use this library https://github.com/sqlcipher/android-database-sqlcipher in android side for encryption purpose and fmdb/sqlcipher for iOS?

radex:
@waqas19921 I don't think it's necessary on iOS:

> On iOS, the database file (like all files) are encrypted. Other apps can't read it.

It would matter only for jailbroken phones with sandboxing disabled, I think.

And yes, encrypting sqlite makes sense, and would be nice if it was an option — but as discussed above, this has consequences on app bundle size and performance. Would you like to do more research on this and contribute to this effort?

y3g0r:
Here I integrated sqlcipher https://github.com/Nozbe/WatermelonDB/compare/master...y3g0r:feature/sqlcipher. It's quick and dirty, but it works for me so far. Maybe it will help somebody out there to get the idea on how to get sqlcipher working with WatermelonDB.

[...]

PR #597, opened 2019-12-20
Encrypted SQLite DB for Android

afiller:
Hi, as promised in #51 hereby I provide a pull request for encrypted databases on android. It's tested (ci:check and in a test project) and working. Nevertheless I want to address some things:

1) It works with or without password, but in all of the cases it's using another SQLite implementation for android. In my tests it works perfect, but I wanted to tell you. 

2) The new library requires to be mentioned in the licenses, see https://www.zetetic.net/sqlcipher/open-source/

I did not change documentation at all. It's up to you where to mention this.

3) The password can be simply set in the SQLiteAdapter:

```
new SQLiteAdapter({
      schema: defaultSchema,
      dbName: 'myDB',
      password: 'ABC123!'
    })
```

If none is given the DB is not encrypted. Quite easy and backwards compatible.

4) I adjusted the iOS methods accordingly to be the same as in Android. The parameter password is given there as well, but simply not used. I think this makes sense for possible future changes also there.

Feel free to change what you want. :) I hope it's a valuable contribution and I would look forward to be mentioned in the contributors section - but of course that's not a "must".

Best regards,

Andreas


alexandrethsilva:
hi @afiller is there any update on this here? anything I could help with? right now it seems to be only a pending conflict res, but maybe there was some back channeling I'm not aware of :)

Overtorment:
@afiller any progress on this?

falquinho:
@afiller Hey, good evening. I'm here to ask for some guidance if you're able to help. My project requires database encryption(late requisite, else I would not have used Watermelon) so i'm trying to port you patch to Watermelon 0.17.1.
It has some differences in the `NativeBridgeType` and `NativeDispatcher`. I've changed their functions signatures where needed to take an extra string for the password, but i keep getting the run-time error `DatabaseBridge.initialize got 5 arguments but requires 6`. It doesn't make sense to me, its args are 5: `tag, schema, password, schemaVersion and callback`.
Can you help make this patch? Can someone point me what I'm missing?

radex:
Closed in favor of https://github.com/Nozbe/WatermelonDB/pull/907 . Thank you for your work!

PR #603, opened 2019-12-31
Fix simple observer bug

kokusGr:
It fixes a bug in simple observation introduced with recent changes. Sometimes after initial emission but before subscribing to observe changes source observable might complete and call cleanup function leading to possible memory leaks.

In NT it only happens on native when using `first` and `toPromise` on observable. I am not sure why. I think it might be because of Promise implementation. However there are probably more edge cases where this bug will occur. 

I haven't added regression tests since I am not sure how to test this properly yet :/ I will try to add one later this week.


radex:
weird but ok.

CI fails, but as I now realize this broke on my PR - not sure how this happened that we didn't notice and merged it anyway :/

PR #605, opened 2020-01-14
Upgrade RN version to build example

jlemm45:
**Summary**
When trying to build the native example, I get the following error. 
Screen Shot 2020-01-14 at 12 45 09 PM

Bumping the RN a bit seems to fix this issue and allows me to build. 

Reference: https://stackoverflow.com/questions/55435943/swift-class-extensions-and-categories-on-swift-classes-are-not-allowed-to-have


radex:
Thanks!

Issue #559, opened 2019-11-08
"@field" decorator give an type error


yardz:
After update 0.15.0 the "field" decorator started to give an error in typescrip.

```
Unable to resolve signature of property decorator when called as an expression.
  This expression is not callable.
    Not all constituents of type 'Object | RawDecorator' are callable.
      Type 'Object' has no call signatures.
```

bogdancochioras-clarisoft:
I also have the issue, it is caused by the fact that the `Descriptor` of the `makeDecorator` is mapped to Object, I managed to resolve that by changing the `Descriptor` definition to:
```
export type Descriptor = any
```
in `/utils/common/makeDecorator/index.d.ts`

ospfranco:
Is there any way to monkey patch this just with typescript?

ospfranco:
nvm ended up using patch-package to modify the types, in `/utils/common/makeDecorator/index.d.ts`:
```
declare module '@nozbe/watermelondb/utils/common/makeDecorator' {
  import { ReplaceReturn } from '@nozbe/watermelondb/utils/common';

  export type Descriptor = any;
  export type RawDecorator = (
    target: Object,
    key: string,
    descriptor: Descriptor,
  ) => Descriptor;
  export type RawDecoratorFactory<T extends any[]> = (
    ...any: T
  ) => RawDecorator;

  export type Decorator<
    Args extends any[],
    Factory extends RawDecoratorFactory<Args>
    // TODO: fix
  > = any;

  export default function makeDecorator<
    Args extends any[],
    T extends RawDecoratorFactory<Args>
  >(): Decorator<Args, T>;
}
```

bogdancochioras-clarisoft:
@ospfranco I did the same thing but there is also an issues with `@readonly` since it wants `@readonly()` an in the case where you have `@readonly @field('aField') aField: string;` not sure if it is correct to replace it with `@readonly()`. One solution is to remove it for know which from what @radex said in the documentation would have probably a minimal performance impact since read-only fields might get optimized.

tomdaniel0x01:
I am also experiencing this error. Looks to be related to #529

tomdaniel0x01:
> nvm ended up using patch-package to modify the types, in `/utils/common/makeDecorator/index.d.ts`:

@ospfranco Wow. This is the first I've heard of patch-package. Perfect! Thanks for sharing.

ScreamZ:
Can we get a fix that's not that hacky ? @radex

[...]

PR #613, opened 2020-01-26
Fix Types

azizhk:
So the original rationale being these types was that they followed flow types.
But certain things are not supported in both. So they might vary from time to time.

🍉 supports both `@nochange fieldName` and `@nochange() fieldName`. Tested both works now.

Fixes https://github.com/Nozbe/WatermelonDB/issues/559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.
 - [LokiJS] Fixed an issue preventing database from saving when using `experimentalUseIncrementalIndexedDB`
 - Fixed a potential issue when using `database.unsafeResetDatabase()`
 - [iOS] Fixed issue with clearing database under experimental synchronous mode
+- [Typescript] Fixed types of decorators.
 
 ### New features (Experimental)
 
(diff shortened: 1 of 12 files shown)


azizhk:
Because the types are not auto generated, was thinking of a test strategy for this. Can we add a "With Typescript" example in the examples folder and then run type check on Travis for that folder.

radex:
> Because the types are not auto generated, was thinking of a test strategy for this. Can we add a "With Typescript" example in the examples folder and then run type check on Travis for that folder.

Yeah! I think you can just make `src/type-tests.ts` or something like that - waiting for the PR :)

radex:
@azizhk can you add a quick note to CHANGELOG about your fixes? then I can merge

azizhk:
> @azizhk can you add a quick note to CHANGELOG about your fixes? then I can merge

Done.

> Yeah! I think you can just make src/type-tests.ts or something like that - waiting for the PR :)

Yeah will do. Will try to make it such that it covers most of the use cases of how to type models. Will try to give PR by 1st week of March.

radex:
thanks!

sidferreira:
We need this deployed...

Issue #612, opened 2020-01-23
Better debugging information for Diagnostic Errors


esbenp:
We are dealing with a lot of `Diagnostic error: Cannot update a record with pending updates` errors these days. It's extremely hard to debug because it happens of course on somebody's client somewhere and is entirely dependant on their local state.

Would probably be easier to debug if table + ID was part of this error - maybe even properties. Something like `Diagnostic error: Cannot update a record with pending updates (table: documents, ID: 1234, pending_updates: title, body)`. Otherwise it's practically impossible to figure out *what caused the error*

What do you think? I could whip out a quick PR

radex:
@esbenp Go for it! I'm a fan of helpful diagnostics :)

esbenp:
Alright I will. I am not super familiar with everything but could you maybe elaborate on what is a common cause for this? Is it `preparedUpdate`s that are never flushed, or is it flushes that fail? Understanding what it means will help with making a better log statement

radex:
this sets the pending update flag:

https://github.com/Nozbe/WatermelonDB/blob/c99f4a3c09354e33ab983b29c1d1be421ab48308/src/Model/index.js#L110

and this resets it:

https://github.com/Nozbe/WatermelonDB/blob/ee81f5ab6a95c5b930fddbe54a4a64bdcd103903/src/Database/index.js#L93

both happen synchronously when you correctly use this:

```
const record = record.prepareUpdate(x)
await database.batch(record)
```

so to hit this bug, you either have to prepare and update and then NOT pass it to batch synchronously (that's a bug, and watermelon in dev mode sends warnings about this), or something crashed on this path. example:

```
await database.batch(
  record1.prepareUpdate(),
  record2.prepareUpdate(x => { throw new Error() } )
)
```

record1 will be stuck with a pending update, because something else crashed before it got to the batch call

esbenp:
This is super helpful information, thank you! I actually thought this was related to multiple clients editing/deleting the same thing and then crashing when syncing (this error appears when we sync in our app).

Searching for this error gave me nothing, but I am thinking this would be extremely helpful as a "Common errors" in docs - what do you say?

radex:
@esbenp yeah! that would be very helpful! `docs-master` is where you can put in .md format documentation

bureyburey:
joining in on the issue.

using these versions:
@nozbe/watermelondb: 0.21.0
@nozbe/with-observables: 1.2.0

[...]

PR #615, opened 2020-01-31
Common mistakes docs

esbenp:
Ref #612 

I gave it a try @radex . I am having a hard time coming up with a great example that demonstrates "do not put await between `prepareUpdate` and `batch`. This is the best I could come up with, but it also includes a logic flaw (since `comments.fetch()` would not include the newly updated comments and therefore `total_good_comments` is wrong). Do you have any feedback for a better example and/or better wordings?

### DIFF
--- a/docs-master/CommonMistakes.md
+++ b/docs-master/CommonMistakes.md
@@ -1,0 +1,135 @@
+# Common Mistakes
+
+This page will try collect some common mistakes and how to avoid them.
+
+## Not ensuring database writes happen in an action
+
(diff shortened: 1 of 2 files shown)


olliewd:
Just wanted to say thanks. I actually found this helpful, and it enabled me to solve a bug I was having. 

I don't think it's super obvious (?) that prepareUpdates can't be called multiple times in the same batch for the same record. Maybe this could be signalled in the docs somehow, or maybe the pull request could be accepted?

PR #617, opened 2020-01-31
better error message when collections are missing

### LINKED ISSUE #616: Warn about missing collections
If you forgot to add a model to `Database` you will get a cryptic typeerror ala

Screenshot 2020-01-31 at 15 07 56

I would suggest some warning in `getAllRecordsToApply` but I am uncertain how to add something like that when using `rambdax`??

```
const getAllRecordsToApply = (
  db: Database,
  remoteChanges: SyncDatabaseChangeSet,
): AllRecordsToApply =>
  piped(
    remoteChanges,
    map((changes, tableName: TableName<any>) => {
      const collection = db.collections.get((tableName: any))

      invariant(collection, `You are trying to sync a collection named ${tableName}, but currently this collection does not exist.` + 
        `Have you remembered to add it to your Database constructor\'s modelClasses property?`)
      
      return recordsToApplyRemoteChangesTo(collection, changes),
    },
    promiseAllObject,
  )
```

something like this?

esbenp:
Closes #616

### DIFF
--- a/src/sync/impl/applyRemote.js
+++ b/src/sync/impl/applyRemote.js
@@ -163,9 +163,16 @@ const getAllRecordsToApply = (
 ): AllRecordsToApply =>
   piped(
     remoteChanges,
-    map((changes, tableName: TableName<any>) =>
-      recordsToApplyRemoteChangesTo(db.collections.get((tableName: any)), changes),
-    ),
+    map((changes, tableName: TableName<any>) => {
+      const collection = db.collections.get((tableName: any))
+
+      if (!collection) {
+        Promise.reject(`You are trying to sync a collection named ${tableName}, but currently this collection does not exist.` + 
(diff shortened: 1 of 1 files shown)


src/sync/impl/applyRemote.js
```diff
@@ -163,9 +163,16 @@ const getAllRecordsToApply = (
...
   piped(
     remoteChanges,
-    map((changes, tableName: TableName<any>) =>
-      recordsToApplyRemoteChangesTo(db.collections.get((tableName: any)), changes),
-    ),
+    map((changes, tableName: TableName<any>) => {
+      const collection = db.collections.get((tableName: any))
+
+      if (!collection) {
+        Promise.reject(`You are trying to sync a collection named ${tableName}, but currently this collection does not exist.` +
```
radex: you forgot to return ;)

radex:
Thanks @esbenp !

PR #622, opened 2020-02-15
Sort take skip query operators

Kenneth-KT:
Added experimental `Q.sortBy(sortColumn, sortOrder)`, `Q.take(count)`, `Q.skip(count)` methods
Add support for `sortBy`, `take` and  `skip` operators into Sqlite adapter

### DIFF
--- a/src/QueryDescription/index.d.ts
+++ b/src/QueryDescription/index.d.ts
@@ -49,10 +49,29 @@ declare module '@nozbe/watermelondb/QueryDescription' {
     left: ColumnName
     comparison: Comparison
   }
-  export type Condition = Where | On
+  export interface SortBy {
+    type: 'sortBy'
+    sortColumn: ColumnName
+    sortOrder: SortOrder
+  }
+  export type SortOrder =
+    | 'asc'
+    | 'desc'
+  export interface Take {
+    type: 'take'
+    count: number
+  }
+  export interface Skip = {
+    type: 'skip'
+    count: number
+  }
+  export type Condition = Where | On | SortBy | Take | Skip
   export interface QueryDescription {
     where: Where[]
     join: On[]
+    sortBy: SortBy[]
+    take?: Take
+    skip?: Skip
   }
 
   export function eq(valueOrColumn: Value | ColumnDescription): Comparison
@@ -71,6 +90,9 @@ declare module '@nozbe/watermelondb/QueryDescription' {
   export function or(...conditions: Where[]): Or
   export function like(value: string): Comparison
   export function notLike(value: string): Comparison
+  export function sortBy(sortColumn: ColumnName, sortOrder?: SortOrder): SortBy
+  export function take(count: number): Take
+  export function skip(count: number): Skip
   export function sanitizeLikeString(value: string): string
 
   type _OnFunctionColumnValue = (table: TableName<any>, column: ColumnName, value: Value) => On
--- a/src/QueryDescription/index.d.ts
+++ b/src/QueryDescription/index.d.ts
@@ -65,7 +65,7 @@ declare module '@nozbe/watermelondb/QueryDescription' {
     type: 'skip'
     count: number
   }
-  export type Condition = Where | On | SortBy | Take | Skip
+  export type Clause = Where | On | SortBy | Take | Skip
   export interface QueryDescription {
     where: Where[]
     join: On[]
@@ -109,7 +109,7 @@ declare module '@nozbe/watermelondb/QueryDescription' {
 
   export const on: OnFunction
 
-  export function buildQueryDescription(conditions: Condition[]): QueryDescription
+  export function buildQueryDescription(conditions: Clause[]): QueryDescription
   export function queryWithoutDeleted(query: QueryDescription): QueryDescription
   export function hasColumnComparisons(conditions: Where[]): boolean
 }
--- a/src/QueryDescription/index.d.ts
+++ b/src/QueryDescription/index.d.ts
@@ -57,11 +57,13 @@ declare module '@nozbe/watermelondb/QueryDescription' {
   export type SortOrder =
     | 'asc'
     | 'desc'
+  export const asc: SortOrder
+  export const desc: SortOrder
   export interface Take {
     type: 'take'
     count: number
   }
-  export interface Skip = {
+  export interface Skip {
     type: 'skip'
     count: number
   }
--- a/src/QueryDescription/index.d.ts
+++ b/src/QueryDescription/index.d.ts
@@ -92,9 +92,9 @@ declare module '@nozbe/watermelondb/QueryDescription' {
   export function or(...conditions: Where[]): Or
   export function like(value: string): Comparison
   export function notLike(value: string): Comparison
-  export function sortBy(sortColumn: ColumnName, sortOrder?: SortOrder): SortBy
-  export function take(count: number): Take
-  export function skip(count: number): Skip
+  export function experimentalSortBy(sortColumn: ColumnName, sortOrder?: SortOrder): SortBy
+  export function experimentalTake(count: number): Take
+  export function experimentalSkip(count: number): Skip
   export function sanitizeLikeString(value: string): string
 
   type _OnFunctionColumnValue = (table: TableName<any>, column: ColumnName, value: Value) => On
--- a/src/QueryDescription/index.js
+++ b/src/QueryDescription/index.js
@@ -53,6 +53,8 @@ export type On = $RE<{
 export type SortOrder =
   | 'asc'
   | 'desc'
+export const asc: SortOrder = 'asc'
+export const desc: SortOrder = 'desc'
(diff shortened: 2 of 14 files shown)


src/QueryDescription/index.d.ts
```diff
@@ -49,10 +49,29 @@ declare module '@nozbe/watermelondb/QueryDescription' {
...
+    | 'desc'
+  export interface Take {
+    type: 'take'
+    count: number
+  }
+  export interface Skip = {
+    type: 'skip'
+    count: number
+  }
+  export type Condition = Where | On | SortBy | Take | Skip
```
radex: do you have an idea how to rename Condition to something that would better describe the extended role?

src/QueryDescription/index.js
```diff
@@ -51,8 +50,30 @@ export type On = $RE<{
   left: ColumnName,
   comparison: Comparison,
 }>
-export type Condition = Where | On
-export type QueryDescription = $RE<{ where: Where[], join: On[] }>
+export type SortOrder =
+  | 'asc'
+  | 'desc'
```
radex: maybe we can export those so that you can use Q.asc, Q.desc and not worry about a mistake?

src/QueryDescription/index.js
```diff
@@ -231,7 +264,32 @@ export const on: OnFunction = (table, leftOrWhereDescription, valueOrComparison)
...
+      case 'skip':
+        // $FlowFixMe: Flow is too dumb to realize that it is valid
+        clauses[type] = cond
+        break
+      default:
+      case 'where':
+        clauses.where.push(cond)
+        break
+    }
+  })
```
radex: I don't get it… aren't you just pushing to `clauses[newType]` in all cases, just that newType has different names? may be easier to have a helper function with a switch  that returns one of `join`, `sortBy`, etc?

src/QueryDescription/index.js
```diff
@@ -253,6 +313,7 @@ export function queryWithoutDeleted(query: QueryDescription): QueryDescription {
   const { join, where: whereConditions } = query
 
   const newQuery = {
+    ...query,
     join: [...join, ...joinsWithoutDeleted(join)],
     where: [...whereConditions, whereNotDeleted],
   }
```
radex: i think you also have to fix up `Query.extend()` -- i guess it would be better if it called a function defined in _this_ file, and not manipulated on `join, where` directly to avoid mistakes in the future

src/QueryDescription/test.js
```diff
@@ -479,6 +506,87 @@ describe('QueryDescription', () => {
...
+      join: [],
+      sortBy: [],
+      take: {
+        type: 'take',
+        count: 100,
+      },
+      skip: {
+        type: 'skip',
+        count: 800,
+      },
```
radex: and also locate the test for Query.extend and add a test for that

src/adapters/sqlite/encodeQuery/index.js
```diff
@@ -151,6 +151,28 @@ const encodeAssociation: (TableName<any>) => AssociationArgs => string = mainTab
 const encodeJoin = (table: TableName<any>, associations: AssociationArgs[]): string =>
   associations.length ? associations.map(encodeAssociation(table)).join('') : ''
 
+const encodeOrderBy = (table: TableName<any>, sortBys) => {
```
radex: add the types

src/adapters/sqlite/encodeQuery/index.js
```diff
@@ -151,6 +151,28 @@ const encodeAssociation: (TableName<any>) => AssociationArgs => string = mainTab
...
+  if (sortBys.length === 0) {
+    return ''
+  }
+  const orderBys = sortBys.map(sortBy =>
+    `${encodeName(table)}.${encodeName(sortBy.sortColumn)} ${sortBy.sortOrder}`
+  ).join(', ')
+  return ` order by ${orderBys}`
+}
+
+const encodeLimitOffset = (take, skip) => {
```
radex: add the types

src/adapters/sqlite/encodeQuery/index.js
```diff
@@ -151,6 +151,28 @@ const encodeAssociation: (TableName<any>) => AssociationArgs => string = mainTab
 const encodeJoin = (table: TableName<any>, associations: AssociationArgs[]): string =>
   associations.length ? associations.map(encodeAssociation(table)).join('') : ''
 
+const encodeOrderBy = (table: TableName<any>, sortBys) => {
+  if (sortBys.length === 0) {
+    return ''
+  }
+  const orderBys = sortBys.map(sortBy =>
+    `${encodeName(table)}.${encodeName(sortBy.sortColumn)} ${sortBy.sortOrder}`
```
radex: can you add an invariant to Q.sortBy to ensure that you can't put something other than asc/desc there?

src/adapters/sqlite/encodeQuery/test.js
```diff
@@ -111,4 +111,39 @@ describe('SQLite encodeQuery', () => {
       `select "tasks".* from "tasks" where "tasks"."col1" like '%abc' and "tasks"."col2" not like 'def%' and "tasks"."_status" is not 'deleted'`,
     )
   })
+  it('encodes order by clause', () => {
+    const query = new Query(mockCollection, [
+      Q.sortBy('sortable_column', 'desc'),
```
radex: can you have multiple sortbys ? if so, add a test for that

src/QueryDescription/index.js
```diff
@@ -198,6 +219,18 @@ export function or(...conditions: Where[]): Or {
...
 
+export function sortBy(sortColumn: ColumnName, sortOrder: SortOrder = 'asc'): SortBy {
+  return { type: 'sortBy', sortColumn, sortOrder }
+}
+
+export function take(count: number): Take {
+  return { type: 'take', count }
+}
+
+export function skip(count: number): Skip {
```
radex: I'd prefer if you call those three functions experimentalSortBy, experimentalTake, experimentalSkip -- the reason why is that I'm not entirely sure if mixing together query conditions and extra parameters like this is the correct API. I don't know what a better API would be, and I don't want to stop you from pushing such an awesome change until it can be perfect -- but I want to set the right expectation to 🍉 users that the exact API may change

kesha-antonov:
Sort would be useful

src/QueryDescription/index.js
Kenneth-KT: The problem is the code is difficult to be statically analyzed for type checking (although type checking for dynamically accessed property as switch condition seems to be broken in Flow, it does work in TypeScript), but maybe the `where` case can be merged into `sortBy` case.

src/QueryDescription/index.d.ts
Kenneth-KT: In SQL, things like `WHERE`, `ORDER BY` and `LIMIT` are collectively called "SQL Clauses".
Maybe call it "QueryClause"? Or "QuerySpecifier"?

radex: Yeah, `Clause` sounds reasonable!

code-brewer:
Hi, all 
Does Take | Skip usable on the production env?  Don't see they are described in the documentation.

radex:
@Kenneth-KT hey, what's up? Do you think you'll have time anytime soon to finish up this -- and the other awesome PR of yours?

src/QueryDescription/test.js
Kenneth-KT: see https://github.com/Nozbe/WatermelonDB/pull/622/files#diff-cc4cc709a149f36c03a3875eb540a4afR84

src/QueryDescription/index.js
Kenneth-KT: refactored to remove repeated push calls

Kenneth-KT:
> @Kenneth-KT hey, what's up? Do you think you'll have time anytime soon to finish up this -- and the other awesome PR of yours?

Hi, although I have resolved nearly most of the code review comments, it seems that more work is needed for the observability handling, will keep pushing new commits if I have time for that 😉 .

radex:
@Kenneth-KT Thank you for your work on this! I'm gonna make my own branch off of yours and try to finish it enough to be mergeable

radex:
@Kenneth-KT BTW: I think I fixed observability (I haven't actually tested this). The point is simply to fall back to the less efficient method that simply re-fetches the query if any change occurs in a relevant table -- same method is used to observe queries with joins.

Kenneth-KT:
> @Kenneth-KT Thank you for your work on this! I'm gonna make my own branch off of yours and try to finish it enough to be mergeable

Thank you. I really appreciate your help.

> @Kenneth-KT BTW: I think I fixed observability (I haven't actually tested this). The point is simply to fall back to the less efficient method that simply re-fetches the query if any change occurs in a relevant table -- same method is used to observe queries with joins.

Great, I think that can be applied to full text search feature as well.

PR #637, opened 2020-03-02
Add Typescript test & example

azizhk:
Followup on #613


examples/typescript/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "typescript-example",
+  "version": "1.0.0",
+  "license": "MIT",
+  "scripts": {
+    "test": "tsc --noEmit"
```
azizhk: Need to add this command to travis tasks. Will need help here.

radex: this is the important bit: https://github.com/Nozbe/WatermelonDB/blob/master/.travis.yml#L31 -- so if you add to the root package.json something like `test:typescript = ... cd examples/... yarn testst` and then to `ci:check` add `previous commands && yarn test:typescript` that śhould work

azizhk: Done.

radex:
@azizhk is this ready for review?

azizhk:
Yes this is ready for review. I'll just squash all commits.

radex:
@azizhk don't squash. please add a note to CHANGELOG

radex:
thanks!

PR #644, opened 2020-03-11
JSI - part 1 - prep

radex:
it'll take me a while to make https://github.com/Nozbe/WatermelonDB/pull/541 reviewable (it's kinda hard to do it incrementally while keeping high CR quality because i first want to make the thing work, and then work on cleaning up, adding error handling, etc...)

so i'm just extracting a few things that have no functional change, but will become important later


native/ios/WatermelonDB/JSLockPerfHack.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#include <jsi/jsi.h>
+
+void watermelonCallWithJSCLockHolder(facebook::jsi::Runtime& rt, std::function<void (void)> block);
```
radex: Note: those two files are exactly the same as in NT, so no need to thoroughly review it

native/ios/WatermelonDB/.clang-format
```diff
@@ -0,0 +1,47 @@
+AccessModifierOffset: 0
```
radex: this is like ESLint but for c++

rozPierog:
@radex something wrong is happening on Travis, will you check it out?

radex:
@rozPierog travis has been fucked on this repo for a while. I don't have the energy right now to fix this. But I run all tests locally before making releases.

PR #648, opened 2020-03-16
Enhancement: Added typescript definition for setGenerator

AndreasA:


azizhk:
Can you add a line in tests as how `setGenerator` is used to 
https://github.com/Nozbe/WatermelonDB/blob/master/examples/typescript/ts-example.ts

AndreasA:
> Can you add a line in tests as how `setGenerator` is used to
> https://github.com/Nozbe/WatermelonDB/blob/master/examples/typescript/ts-example.ts

Done. Added it to exmples

radex:
thanks!

PR #664, opened 2020-04-21
replace the sql parser on android to fixed a potential crash

xelven:
replace the sql parser from string to fixed a crash issue when it try to parser a mass sql script string if it contian a grammar such as 'create trigger' with double ';'

currently is parser sql with ";" is too bad, so I create my own sql parser base on protocol: http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt

otherwise it will 100% crash when run :
`create trigger "aaas_fts_delete" after delete on "aaas" begin delete from "aaas_fts" where "rowid" = OLD.rowid; end;`

### DIFF
--- a/native/android/build.gradle
+++ b/native/android/build.gradle
@@ -22,6 +22,7 @@ dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation 'com.github.xelven:SqlParser-Android:1.0'
 }
 repositories {
     mavenCentral()
--- a/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
@@ -3,6 +3,7 @@ package com.nozbe.watermelondb
 import android.content.Context
 import android.database.Cursor
 import android.database.sqlite.SQLiteDatabase
+import org.xelven.sqlparser.SQLParser
 import java.io.File
 
 class Database(private val name: String, private val context: Context) {
@@ -27,8 +28,11 @@ class Database(private val name: String, private val context: Context) {
 
     fun executeStatements(statements: SQL) =
             transaction {
-                statements.split(";").forEach {
-                    if (it.isNotBlank()) execute(it)
+                var parser = SQLParser()
+                parser.setUseScriptDetecting(true)
+                parser.parseScript(statements)
+                parser.getStatements().forEach {
+                    if (it.sql.isNotBlank()) execute(it.sql)
                 }
             }
 


radex:
can you explain in more detail why the current implementation causes the crash? as in, what's the root cause?

I don't think we're going to accept an addition of a 3rd party dependency -- you'll have to show why there's no reasonable workaround for the issue

Kenneth-KT:
One background information is Android's `execSql` method cannot accept multiple SQL statements within one string, delimited by semi-colons `;`, but iOS's counterpart could.
So `statements.split(";")` is written to achieve "executing multiple SQL statements by passing one string".
However, simply splitting string without context (ie. understanding each character's role precisely) is very dangerous and error prone, you could accidentally misinterpret a valid SQL statement into multiple invalid statements.
For example:
```sql
select "tasks".* from "tasks" where "tasks"."description" is 'too long; didnt read'
```
would become two statements:
1:
```sql
select "tasks".* from "tasks" where "tasks"."what" is 'too long;
```
2:
```sql
 didnt read'
```
which definitely cannot be executed.

Therefore, I can think of two choices:
1. implement a proper (context sensitive) SQL parser to interpret multi-statements SQL string correctly
2. forbid the use of multi-statements SQL strings, either by disabling such handling for iOS SQL executor or simply stating that clear in documentations, at the same time provide another API for executing multiple statements in one call as array of string.

@radex how about that?

radex:
> However, mindlessly splitting SQL strings without grammatical context (ie. understanding role of each character precisely) is very dangerous and error prone, you could accidentally misinterpret a valid SQL statement into multiple invalid statements.

Of course! I get that. But `executeStatements` is not a public api, it's only used for:

```
database.executeStatements(schema.sql + Queries.localStorageSchema)
database.executeStatements(migrations.sql)
```

which are generated by Watermelon functions that don't seem to allow dangerous/invalid SQL to be generated. So what's the actual case where this would fail?

> implement a proper (grammar sensitive) SQL parser to interpret multi-statements SQL string correctly, as in this PR attempted to do

that's waaaaay over-engineered for such a minor use case

> refactor executeStatements into accepting multi-statements SQL via array of string, instead of one string of multi-statements SQL delimited by semi-colons

that's more like it, would require encodeSchema and encodeMigrations to be refactored.

but besides, unless there's a very general use case (that's not a one of hack in your app you can work around), I think it may be a waste of time, as we'll be moving towards a JSI-based database adapter that doesn't have this problem (https://github.com/Nozbe/WatermelonDB/blob/master/native/shared/Database.cpp#L128-L144)

xelven:
@radex 
the actual case happened when u merged my colleague @Kenneth-KT 's supported `Full Text Search` PR, it 100% crashed on android when initial full text search feature with sql `create trigger`

radex:
Understood. That makes sense to me.

So, as I understand, you have your own patch that works for you. I'd rather not merge a proper SQL parser just for this. I'd be leaning towards not doing anything about it and supporting FTS only in JSI mode (once that's properly merged). But refactoring encodeSchema, encodeMigrations to return an array of strings is also alright.

xelven:
yep, I have my own patch work for me, ya I think is also good to refactoring encodeSchema.
I was want to solve it more thoroughly, so implemented sql parser lib for it haha

Kenneth-KT:
> Understood. That makes sense to me.
> 
> So, as I understand, you have your own patch that works for you. I'd rather not merge a proper SQL parser just for this. I'd be leaning towards not doing anything about it and supporting FTS only in JSI mode (once that's properly merged). But refactoring encodeSchema, encodeMigrations to return an array of strings is also alright.

Yeah, I think refactoring it to only accept string array is the optimal solution too.
However I just wonder if it could break any plugins/extensions or other things that depends on it? Or should I let user pass in both string or string array, then at runtime, determine what behavior to use, string for old, string array for new behavior?

radex:
> However I just wonder if it could break any plugins/extensions or other things that depend on it? Or, for backward compatibility, should we let user pass in either string or string array, then at runtime, determine what behavior to use: string for old, string array for new behavior?

No, it's an implementation detail, not a public API.

Issue #653, opened 2020-03-27
Flipper support


ScreamZ:
Hey, I would like to add Flipper support and I followed this tutorial:

https://fbflipper.com/docs/setup/databases-plugin.html

But I need the database path, could you help me? Thanks

kesha-antonov:
Interested too

kesha-antonov:
@ScreamZ 
Did you build RN 0.62 with WatermelonDB?
It gives me errors

cesargdm:
@kesha-antonov Getting errors too

kesha-antonov:
Without Flipper it builds

zecakeh:
It builds for me with Flipper. To make WatermelonDBs' database discoverable, I followed the docs and replaced:

```java
databaseFiles.add("...path_to_your_db...")
```
With:

```java
String watermelondb = context.getDatabasePath("watermelon.db").getPath().replace("/databases", "");
databaseFiles.add(new File(watermelondb));
```

kilbot:
@kesha-antonov @cesargdm  There seems to be a [conflict with Flipper on iOS](https://github.com/Nozbe/WatermelonDB/issues/655#issuecomment-610350915), but you can use Flipper to debug your Android app using the method given by @ScreamZ.

Your  `ReactNativeFlipper.java` should look something like this - https://gist.github.com/kilbot/253e4bfb0b931f9dfc23a45f71564bf7 - where `path_to_your_db` should match the `dbName` passed into the SQLiteAdapter (`watermelon` is the default).

Question for @zecakeh: is it possible to loop through multiple database files?

zecakeh:
@kilbot Of course, just add as many `databaseFiles.add(new File("path/to/database.db"));` as needed. The code in the docs includes both databases that are in the context and a custom one.

inferusvv:
> Your `ReactNativeFlipper.java` should look something like this - https://gist.github.com/kilbot/253e4bfb0b931f9dfc23a45f71564bf7 - where `path_to_your_db` should match the `dbName` passed into the SQLiteAdapter (`watermelon` is the default).

This is something that should be added to documentation :)

cheehieu:
[...]

PR #670, opened 2020-04-28
Support flipper on Android

leninlin:
related issue #653


radex:
@rozPierog can you review this?

leninlin:
Any updates?

rozPierog:
Hey @leninlin sorry about the delay, will test it out and review your PR this week. Thanks for contributing!

rozPierog:
Everything seems good! Thanks @leninlin!

rozPierog:
@leninlin After looking at it again I had to revert that merge. Your fix is great but it might break some databases on Watermelon upgrade as place from where database is opened/created might be different between your code and previous implementation. Please ensure that your change won't break existing apps and submit PR again.

leninlin:
@rozPierog thank you for your review. I agree that this is not a good solution.
I will think about this and do a new PR later

PR #671, opened 2020-05-05
Safer reset database

radex:
First step in removing "unsafe" from unsafeResetDatabase. I was planning to do more, but I have some doubts as for the best way to implement it, so I'm committing what's done, and I'll linger on the problem a little more until I have the final solution.

### DIFF
--- a/src/Database/index.js
+++ b/src/Database/index.js
@@ -6,7 +6,7 @@ import { startWith } from 'rxjs/operators'
 import { values } from 'rambdax'
 
 import { type Unsubscribe } from '../utils/subscriptions'
-import { invariant, logger } from '../utils/common'
+import { invariant } from '../utils/common'
 import { noop } from '../utils/fp'
 
 import type { DatabaseAdapter, BatchOperation } from '../adapters/type'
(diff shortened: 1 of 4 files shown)


src/Database/index.js
```diff
@@ -175,14 +177,37 @@ export default class Database {
...
+      // Kill ability to call adapter methods during reset (to catch bugs if someone does this)
+      const { adapter } = this
+      const ErrorAdapter = require('../adapters/error').default
+      this.adapter = (new ErrorAdapter(): any)
+
+      // Check for illegal subscribers
+      if (this._subscribers.length) {
+        // eslint-disable-next-line no-console
+        console.error(
+          'Application error! 1 Database subscriber was detected during database.unsafeResetDatabase() call. App should not hold onto subscriptions or Watermelon objects while resetting database.',
```
kokusGr, line 194: Will it always be `1`? Or should it be `{this._subscribers.length}`?

radex: Oof. Brain fart — thanks for catching this.

kokusGr: I think you need to change it in 1 of the tests as well.

radex: What do you mean? https://github.com/Nozbe/WatermelonDB/pull/671/files#diff-1c7bf3678f7379f801755def5d7c2facR109 is correct - checks if there'e exactly 1 subscriber too many

kokusGr: Right, that's true. CI is failing and that was my first guess.

radex:
ah, yes, travis is broken on this repo - i tested locally though and it's ok ¯\_(ツ)_/¯

kokusGr:
@radex Ok, I am merging then. Thanks :)

PR #672, opened 2020-05-06
Fix typescript type of changes key

kotielnikov:
Inside of js:
```
export type CollectionChange<Record: Model> = { record: Record, type: CollectionChangeType }
export type CollectionChangeSet<T> = CollectionChange<T>[]
...
 changes: Subject<CollectionChangeSet<Record>> = new Subject()
```

**Typescript type should be also an array of `CollectionChange` and not just one `CollectionChange`**


radex:
Thanks!

PR #676, opened 2020-05-13
Sync.md: add tutorial how to build sync backend

fahrinh:
I've  published a post how to build WatermelonDB sync backend in Elixir.
It may provides a starting point for others  to build their own backend.
Feedback is welcome.


radex:
@fahrinh This is wonderful! I've seen and read your article and it's really nice work! I like in particular that you did go a different route than we did with versioning (counting instead of timestamps), but actually took care to try to engineer it correctly :)

fahrinh:
@radex WatermelonDB is easy to use compared to other alternatives. It's game changer. Thanks for your amazing work 👍

PR #681, opened 2020-05-14
fixes #680

### LINKED ISSUE #680: [Typescript] Relation / type consistency
When we look at Relation definition 

```
export default class Relation<T extends Model> {
    public constructor(
      model: Model,
      relationTableName: TableName<T>,
      columnName: ColumnName,
      options: Options,
    )

    public id: $Call<ExtractRecordId<T>>

    public fetch(): Promise<T>

    public set(record: T): void

    public observe(): Observable<T>
  }
```

fetch (same for Observe) returns a Promise of <T>. Actually, when we look at the js code it may be T or null 

I will propose a pull request to change this

mlecoq:
see #680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - [Typescript] Add Tests to test Types.
 - Fixed typo in learn-to-use docs.
 - [Typescript] Fixed types of changes.
+- [Typescript] Fixed types of Relation
 
 ## 0.16 - 2020-03-06
 
--- a/src/Relation/index.d.ts
+++ b/src/Relation/index.d.ts
@@ -21,10 +21,10 @@ declare module '@nozbe/watermelondb/Relation' {
 
     public id: $Call<ExtractRecordId<T>>
 
-    public fetch(): Promise<T>
+    public fetch(): Promise<T|null>
 
     public set(record: T): void
 
-    public observe(): Observable<T>
+    public observe(): Observable<T | null>
   }
 }
--- a/src/Relation/index.d.ts
+++ b/src/Relation/index.d.ts
@@ -7,6 +7,11 @@ declare module '@nozbe/watermelondb/Relation' {
   type ExtractRecordIdOptional<T extends Model> = (value: T | void) => RecordId | void
   type ExtractRecordId<T extends Model> = ExtractRecordIdNonOptional<T> & ExtractRecordIdOptional<T>
 
+
+  type RecordNonOptional<T extends Model> =  T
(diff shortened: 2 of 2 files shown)


radex:
This is better than what was before, but not a complete fix. Notice that now, those fields will ALWAYS be nullable, even if on a non-nullable relation. In Flow there's some trickery to, on type level, figure out if Relation methods/properties should be nullable or not. I bet you can do something similar in TypeScript:

https://github.com/Nozbe/WatermelonDB/blob/master/src/Relation/index.js#L13-L15

mlecoq:
@radex I get your point, but if I look into existing typescript definition for ExtractRecordId

```
type ExtractRecordIdNonOptional<T extends Model> = (value: T) => RecordId
type ExtractRecordIdOptional<T extends Model> = (value: T | void) => RecordId | void
type ExtractRecordId<T extends Model> = ExtractRecordIdNonOptional<T> & ExtractRecordIdOptional<T>
```

the result is the same as 

```
type ExtractRecordId<T extends Model> = (value: T | void) => RecordId | void
```

I have currently no way to determine if Relation is optional or not 

to be consistent, I can change add 

```
  type RecordNonOptional<T extends Model> =  T
  type RecordOptional<T extends Model> =  T | null
  type Record<T extends Model> = RecordNonOptional<T> | RecordOptional<T>
```

and then 

```
    public fetch(): Promise<Record<T>>

    public set(record: Record<T>): void

    public observe(): Observable<Record<T>>
```

But the result will still be that `observe()` and `fetch()` will return `T` or `null` (regardless it is optional or not) 

Does that suit to you ? Or I didn't get your message ?

zhuSass:
The official has helped us to configure the type, can be introduced
```
import { default as Relation } from '@nozbe/watermelondb/Relation';
@relation('exercises', 'exercise_id') public exercise!: Relation<
    ExerciseModel
  >;
```

radex:
I'm not familiar enough with TypeScript to make the right call here. @jakebloom hey, you contributed TS types in the past - can you take a look at this?

jakebloom:
This makes sense to me but I'm not sure why we need the extra type `Record` (and the related optionality). We can just do `T | null` if we want to go for consistency with the js code. 

We should probably modify the flow types to indicate nullability here too

mlecoq:
Right, I will update my PR

radex:
@mlecoq is this done?

mlecoq:
@radex yes, I have modified the type as suggested by @jakebloom 

see https://github.com/Nozbe/WatermelonDB/pull/681/commits/5520450b66b374baf7fd9858fd0fb99471e1faf5

Issue #691, opened 2020-05-19
[Bug][iOS] 'jsi/jsi.h' file not found


kesha-antonov:
Hi!

After update from `0.16.0` to `0.16.1` I get this error:
```
In file included from /node_modules/@nozbe/watermelondb/native/ios/WatermelonDB/JSLockPerfHack.mm:1:

/node_modules/@nozbe/watermelondb/native/ios/WatermelonDB/JSLockPerfHack.h:3:10: fatal error: 'jsi/jsi.h' file not found
#include <jsi/jsi.h>
         ^~~~~~~~~~~
1 error generated.
```

<img width="946" alt="Screenshot 2020-05-19 at 13 03 05" src="https://user-images.githubusercontent.com/11584712/82313646-4f8bd480-99d1-11ea-9746-30346a05281f.png">

radex:
@kesha-antonov RN version? are you using cocoa pods or classic Xcode projects?

kesha-antonov:
@radex 

"react-native": "0.63.0-rc.1"

I'm using Pods

radex:
Okay. I will investigate sometime soon. Sorry about this - for now, just use 0.16.0 ;)

kesha-antonov:
> Okay. I will investigate sometime soon. Sorry about this - for now, just use 0.16.0 ;)

thanks!

hussamkurd:
Same here

hussamkurd:
@radex 
even with 0.16.0
<img width="597" alt="Screen Shot 2020-05-22 at 11 02 33 AM" src="https://user-images.githubusercontent.com/15825449/82645652-e21fb400-9c1b-11ea-8881-6aaaa244473c.png">

thiagobrez:
Same here, downgrading to 0.16.0 works.

`react-native`: 0.61.5
Using pods

zeo1986:
have the same error with 0.16.0 =(((

mtford90:
Yep - seems to effect >=0.16.

Had to drop down to 0.15 here.

iamclaytonray:
@mtford90 - would you happen to have a repo that you could make public? I've struggled all morning to get this lib running properly 😅

mtford90:
@iamclaytonray I don't have a public repo I can show you but I've had a rather wonky experience getting this library running myself.

0.15.0
- Runs in dev mode, doesn't run in release mode

0.16.0,0.16.1
- Will not run in either debug mode or release mode

[...]

PR #704, opened 2020-05-28
[ios] Add the path to the jsi headers - issue #691

victorbutler:
I added a Header Search Path that contained the JSI headers that the compiler was complaining about.

There is another set of JSI headers inside `react-native/ReactCommon` that seems more appropriate, but other issues came up after added those.

https://github.com/Nozbe/WatermelonDB/issues/691

Please merge if you see fit, or deny if there's a better way to include these headers. I have no idea why they're not present anymore in the header search paths included. Maybe JSI was moved in React Native?


radex:
@victorbutler No, I think your fix is correct - they DID move the JSI headers in 0.62 I think

radex:
Thank you!

PR #708, opened 2020-05-31
Fix crash when syncing big data

leninlin:
Hi!
I got an error while syncing big data. (over 100k)
**"RangeError: Maximum call stack size exceeded"**

screenshot

### DIFF
--- a/src/sync/impl/applyRemote.js
+++ b/src/sync/impl/applyRemote.js
@@ -250,7 +250,7 @@ export default function applyRemoteChanges(
         ? unsafeBatchesWithRecordsToApply(db, recordsToApply, sendCreatedAsUpdated, log)
         : [
             db.batchArray(
-              ...prepareApplyAllRemoteChanges(db, recordsToApply, sendCreatedAsUpdated, log),
+              prepareApplyAllRemoteChanges(db, recordsToApply, sendCreatedAsUpdated, log),
             ),
           ]),
     ])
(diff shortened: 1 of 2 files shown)


src/sync/impl/applyRemote.js
```diff
@@ -249,8 +249,8 @@ export default function applyRemoteChanges(
       ...(_unsafeBatchPerCollection
         ? unsafeBatchesWithRecordsToApply(db, recordsToApply, sendCreatedAsUpdated, log)
         : [
-            db.batch(
-              ...prepareApplyAllRemoteChanges(db, recordsToApply, sendCreatedAsUpdated, log),
+            db.batchArray(
```
radex, line 252: explain - why would this have a significant impact on call stack size?

leninlin: Honestly, I still don’t understand why the error is about the size of the call stack, but this error appears because too many arguments are passed to the function.

radex: Ah! That makes sense, actually - it's not about the depth of the call stack, but the amount memory allocated to the stack.

radex:
@leninlin do you want to add a note to CHANGELOG about this fix? I can merge.

radex:
ok, no need to worry about this anymore - I pulled your changes into https://github.com/Nozbe/WatermelonDB/pull/729, credited you, and also improved - there's only one batch() function, but it can now take a single array as an argument

PR #714, opened 2020-06-03
Migration syncs — Part 1

radex:


radex:
No actual migration sync implemented, just some helpers for it, auxiliary changes, and rewritten Sync documentation

- [DX] Intializing Database with invalid model classes will now show a helpful error
 - [DX] DatabaseProvider shows a more helpful error if used improperly
 - [Sync] Sync no longer fails if pullChanges returns collections that don't exist on the frontend - shows a warning instead. This is to make building backwards-compatible backends less error-prone
 - [Sync] [Docs] Sync documentation has been rewritten, and is now closer in detail to a formal specification
 - [Hardening] database.collections.get() better validates passed value

PR #728, opened 2020-06-17
NodeJS client

### LINKED ISSUE #721: Node JS client
Hi! We use some command-line tools to help with our project.
At this moment we use a workaround as `LokiJS` is `web` and `sqlite` has native dependencies.

I tried to write my own adapter but I couldn't write it properly as we had no access to some parts of code through import.

At first glance, I believe `sqlite` can be modified in order to handle both `native` and `node`, but before starting this PR, I would like to know if it sounds interesting for the maintainers.

sidferreira:
Fix #721 

> note: I wanted to have the js files in the `./native` folder, but as it is out of the `./src` folder, I decided to change. 

pending tests, suggestions welcome
cc @radex


radex:
@sidferreira wow, amazing work! it will take me a while to go through this, so bear with me, but nice.

re: tests. check out `commonTests` in adapters. There's a giant adapters test that runs for all adapter implementations (LokiJS, Android, iOS async, iOS sync, iOS JSI). Take a look at concrete tests for lokijs/ and sqlite/ - they just adapt the commonTests format to the concrete test runner. So it should be very easy for you to hook up a full test suite for your adapter. And I think you'll likely find some bugs caught by those tests.

sidferreira:
@radex I tried the `commonTests` but I'm facing issues.

For example, the test `can query and count on empty db` seems to fail because the table doesn't exist. 

I'm still trying to understand all tests ran and what it is expecting... sometimes it seems to have an empty DB when if should have data...

radex:
@sidferreira look at this bit:

https://github.com/Nozbe/WatermelonDB/blob/master/src/adapters/sqlite/integrationTest.js#L15-L23

I'm passing to tests an initialized adapter, passing `testSchema`. So the table should exist. Or if it actually doesn't, seems like a bug on your side... let me know what you figure out

sidferreira:
@radex yeah, shame on me, I forgot to finish the batch methods. Addressing tests now

sidferreira:
@radex So, after some work, everything seems to be working fine. There are 2 errors happening, but both show the following note: `if this fails, it means the issue's been fixed`. So I think it is a win :)

I made the tests run against the 4 scenarios:
- SQLiteAdapterNode (Synchronous/File)
- SQLiteAdapterNode (Synchronous/Memory)
- SQLiteAdapterNode (Asynchronous/File)
- SQLiteAdapterNode (Asynchronous/Memory)

The shared cache required a small workaround (https://github.com/sidferreira/WatermelonDB/blob/788454f867e203532caa9c5851cb7baf42b83f71/src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js#L74) but required no changes on tests :D

I'm still not happy with the place where the files are (`src/adapters/sqlite/node` but I had no better idea)

Also, I believe the typing may be improved, but I'm new to Flow...

sidferreira:
@radex Oh! important to mention that, because of `fsevents`, I had to force my node to be version `10`. I added a `.nvmrc`file to help but I'm not sure if it is desirable.

radex:
> I made the tests run against the 4 scenarios:

I don't want the full test suite running against the file system, it's going to be really slow on CI. I suggest you add a single test running against the FS (in your suite, not commonTests), to check if it actually works (something simple, open database, batch create a record, open again, see if the record is present), but for the rest we'll trust that the node sqlite dependency is set up correctly for fs, and we don't need to check it.

> The shared cache required a small workaround (https://github.com/sidferreira/WatermelonDB/blob/788454f867e203532caa9c5851cb7baf42b83f71/src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js#L74) but required no changes on tests :D

nice

src/adapters/sqlite/test.js
```diff
@@ -0,0 +1,52 @@
...
+import SqliteAdapter from './index'
+import DatabaseAdapterCompat from '../compat'
+
+function removeIfExists(file, dbName) {
+  if (file && fs.existsSync(dbName)) {
+    fs.unlinkSync(dbName)
+  }
+}
+
+describe.each([
```
radex, line 14: huh, TIL that `describe.each()` is a thing

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,281 @@
...
+      this.database = DatabaseDriver.sharedMemoryConnections[dbName]
+    }
+  }
+
+  find = (table: string, id: string) => {
+    if (this.isCached(table, id)) {
+      return id
+    }
+
+    const query = `SELECT * FROM ${table} WHERE id == ? LIMIT 1`
```
radex, line 88: check for the current version of database drivers on master. There's been some extra escaping added -- I actually thing this shouldn't pass tests, unless you're not branching off of master

src/adapters/sqlite/makeDispatcher/node/Database.js
```diff
@@ -0,0 +1,129 @@
+// @flow
+const fs = require(`fs`)
+const SQliteDatabase = require('better-sqlite3')
+
+class Database {
```
radex, line 5: don't call it `Database`, since that's already a class name for something else (it's not _technically_ a conflict, but it may be confusing to someone in the future)

src/adapters/sqlite/makeDispatcher/index.js
```diff
@@ -0,0 +1,140 @@
...
+    ]
+  })
+
+  const dispatcher = fromPairs(methods)
+  // $FlowFixMe
+  return dispatcher
+}
+
+const initializeJSI = () => {
+  if (global.nativeWatermelonCreateAdapter) {
```
radex, line 98: there's a bunch of junk like this function that really makes no sense in the context of a node implementation.

package.json
```diff
@@ -103,6 +103,7 @@
     "babel-plugin-minify-replace": "^0.5.0",
     "babel-plugin-module-resolver": "^3.1.1",
     "babel-plugin-transform-remove-console": "^6.9.4",
+    "better-sqlite3": "^7.1.0",
```
radex, line 106: you _could_ add it to peerDependencies. Or maybe not.

Well, it would be nice if you could update the docs-master/ folder to mention that SQLite on NodeJS is a thing + a CHANGELOG note about your addition (making sure to mention your @name, of course)

.nvmrc
```diff
@@ -0,0 +1 @@
+10
```
radex, line 1: hmm, i'd rather not, IIUC this selects Node v=10, right? if so, i don't want to do it, as v=11 works just as fine

radex:
I took a quick look at your PR. I think some cleaning up needs to be done, but if it passes tests, I think we're close :)

I'm not sure how I feel about your approach of very very closely replicating Swift code in JS, since in Node.JS, since we don't have to deal with React Native's NativeModules, it could have been much simplified. But never mind - it made your work easier and maybe it will make it easier to maintain (since it's a replica, not a reimplementation), so it's fine.

sidferreira:
@radex indeed. I decided to replicate it to save time, as I wouldn't need to learn how it work. Also the goal is to have the code used on app and on node as similar as possible, including all queries.

If you want we can schedule a call to review and explain some decisions (like the fix args on batch method)

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: I'll rebase, just in case...

src/adapters/sqlite/makeDispatcher/node/Database.js
sidferreira: Again, just following Swift :/

.nvmrc
sidferreira: It's just that I was using 12 locally

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: I checked and it was missing on that one. Also, the tests were working OK on `master`.

src/adapters/sqlite/makeDispatcher/node/Database.js
sidferreira: I just checked and Android uses the same structure...

src/adapters/sqlite/makeDispatcher/index.js
sidferreira: cleaned it up

package.json
sidferreira: yeah, I was thinking about `peerDependencies` too... up to you...

.nvmrc
sidferreira: changed to 11 and works OK

package.json
sidferreira: in the end, I moved to `peerDependency` so we reduce the amount of download if the person won't use it.

Also, it is possible to have an `index.web.js` to proxy `LokiJS` inside of it... making it more transparent...

sidferreira:
Hey @radex, any idea when we'll be able to review this?

BTW, thanks for #731!!!

sidferreira:
Rebased + fixed flow errors.
Had to change the dependency to `devDependencies` to be sure it installs on CI.
Will have to check the best approach to handle it.

radex:
I'll do my best to check it next week :)

dpnolte:
I've checked out your PR because I also want to pre-populate the database in node.
For what's it worth, it also works in my setup with about 1000 records.
I noticed one small difference. With the native adapter, the _db_ extensions gets automatically added.  With the node adapter, it does not and you need to specify the extension yourself.
Thank you all for your great work :)

sidferreira:
@dpnolte that's a great catch! I will address a small fix for that...

sidferreira:
@dpnolte if you have time, please check if this fixes your issue

dpnolte:
@sidferreira  thanks for the update! I've checked and now the `.sqlite` extension gets added.

This might be debatable but I was kinda expecting the extension to be `.db` as that is consistent with how the database file is named in the native adapter ([Android](https://github.com/Nozbe/WatermelonDB/blob/master/native/android/src/main/java/com/nozbe/watermelondb/Database.kt#L20, [iOS](https://github.com/Nozbe/WatermelonDB/blob/928d6e3711be998b9cfffdac09a2f783713ef96f/native/ios/WatermelonDB/DatabaseDriver.swift#L46))

I also noticed that the path always gets prefixed with `process.cwd()`.
This line (https://github.com/sidferreira/WatermelonDB/commit/10c1620884b6b68660f17287942e6f920e0013d9#diff-7a16159c935323ef63ca606211ed002eR45) should probably be changed in an or statement:
``dbName.startsWith('/') || dbName.startsWith('file:') ? dbName : `${process.cwd()}/${dbName}` ``

Thanks for the updates. I hope this will be able to land soon as it a lot of fun and useful to pre-seed the database.

sidferreira:
@dpnolte my mistake the sqlite extension. I was on a hurry! 😂

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+  setUpWithMigrations = (dbName: string, migrations: Migrations) => {
+    this.init(dbName)
+    this.migrate(migrations)
+    this.isCompatible(migrations.to)
+  }
+
+  init = (dbName: string) => {
+    this.database = new Database(getPath(dbName))
+
+    const isSharedMemory = dbName.indexOf('mode=memory') > 0 && dbName.indexOf('cache=shared') > 0
```
radex, line 85: you know that `String.includes` is widely supported, right? ;)

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+    if (isSharedMemory) {
+      if (!DatabaseDriver.sharedMemoryConnections[dbName]) {
+        DatabaseDriver.sharedMemoryConnections[dbName] = this.database
+      }
+      this.database = DatabaseDriver.sharedMemoryConnections[dbName]
```
radex, line 90: does `better-sqlite3` not support shared cache memory mode? I don't believe that this replicates the semantics exactly. But since it passes tests, it's probably not a big deal (since I can't think of many legitimate use cases for using this other than FOR tests), but just saying.

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    this.markAsCached(table, id)
+    return results[0]
+  }
+
+  cachedQuery = (table: string, query: string): any[] => {
+    const results = this.database.queryRaw(query)
+    return results.map((row: any) => {
+      const id = `${row.id}`
```
radex, line 113: why is this wrapped in a `${}` ? Can `row.id` be something other than a string?

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+          case 'destroyPermanently': {
+            const [id] = rest
+            // TODO: What's the behavior if nothing got deleted?
+            this.database.execute(`DELETE FROM '${table}' WHERE id == ?`, [id])
+            removedIds.push([table, id])
+            break
+          }
+
+          default: {
+            break
```
radex, line 164:
```suggestion
            throw new Error('unreachable')
```

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  removeLocal = (key: string) => {
+    this.database.execute('DELETE FROM `local_storage` WHERE `key` == ?', [key])
+  }
+
+  // MARK: - Record caching
+
+  hasCachedTable = (table: string) =>
+    Object.prototype.hasOwnProperty.call(this.cachedRecords, table)
```
radex, line 218: it's OK, but it would be better to use a Map for this to avoid having to do this dance

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
```diff
@@ -0,0 +1,292 @@
...
+
+  markAsCached = (table: string, id: string) => {
+    if (!this.hasCachedTable(table)) {
+      this.cachedRecords[table] = new Set()
+    }
+    this.cachedRecords[table].add(id)
+  }
+
+  removeFromCache = (table: string, id: string) => {
+    if (this.cachedRecords[table] && this.cachedRecords[table].has(id)) {
```
radex, line 235:
```suggestion
    if (this.hasCachedTable(table) && this.cachedRecords[table].has(id)) {
```

src/adapters/sqlite/makeDispatcher/node/DatabaseBridge.js
```diff
@@ -0,0 +1,374 @@
...
+      reject(`db.${functionName}.error`, error.message, error)
+    } else {
+      throw new Error(`db.${functionName} missing reject (${error.message})`)
+    }
+  }
+}
+
+const databaseBridge = new DatabaseBridge()
+
+export default databaseBridge
```
radex, line 374: TBH I don't understand why you went through all the pain and effort to translate Swift `async` and `synchronous` methods given that there is no distinction in node and all methods are synchronous anyway. But OK, I'll accept it.

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: I'm trying to remember WHY I did this instead of `map`. I personally hate this approach.

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: I think it's just cause I got used to... we're migrating from int to string on our project so it's automatic...

src/adapters/sqlite/makeDispatcher/node/Database.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  unsafeDestroyEverything = () => {
+    // Deleting files by default because it seems simpler, more reliable
+    // And we have a weird problem with sqlite code 6 (database busy) in sync mode
+    // But sadly this won't work for in-memory (shared) databases, so in those cases,
+    // drop all tables, indexes, and reset user version to 0
+
+    if (this.isInMemoryDatabase()) {
+      const results = this.queryRaw(`SELECT * FROM sqlite_master WHERE type = 'table'`)
```
radex, line 87: this should be in the transaction

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: So, as far as I spoke to them, no... when in-memory mode, each DB connection seems to be unique. ( https://github.com/JoshuaWise/better-sqlite3/issues/417 )

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
sidferreira: hahaha yeah, I'm getting used to replace it... I just think that `indexOf` is faster (no tests ever made)

CHANGELOG.md
```diff
@@ -75,6 +75,9 @@ Another WatermelonDB release after just a week? Yup! And it's jam-packed full of
        however this speedup is only achieved with some unpublished React Native patches.
 
        To try out JSI, add `experimentalUseJSI: true` to `SQLiteAdapter` constructor.
+- [NodeJS] Introducing a new SQLite Adapter based integration to NodeJS. This requires a
```
radex, line 78: can you please move this line to CHANGELOG-Unreleased.md?

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
radex: weird, but okay

src/adapters/sqlite/makeDispatcher/node/DatabaseBridge.js
sidferreira: Easy: I had no idea on HOW it gonna behave, I needed it working fast. So I did it "duplicated" and later I would improve it as needed. It was much more `translation` than `coding something new`

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
radex: yeah, me too. Last time I checked, it wasn't even faster in microbenchmarks

src/adapters/sqlite/makeDispatcher/node/DatabaseBridge.js
radex: fair enough

src/adapters/sqlite/makeDispatcher/node/DatabaseDriver.js
radex: well it can't hurt, and I'm not going to waste my or your time trying to make this the best possible implementation we could get ;)

src/adapters/sqlite/makeDispatcher/node/Database.js
sidferreira: Mmm... this also means we should have a test failing here... will work on it...

radex:
@sidferreira there's been some test failing. can you take a look? if you don't know how to fix it, just skip the test for node sqlite adapter (writing to a console that the test is broken)

sidferreira:
@radex Some of the tests that were failing seems to be a false negative... I was confused as well. Will rebase, run everything again and fix your comments hopefully today.

src/adapters/sqlite/makeDispatcher/node/Database.js
radex: ideally, yeah, but it might be hard to write a test for ;)

radex:
@sidferreira Thanks! I reviewed all the code, so after fixes, I should be able to just hit merge tomorrow :)

sidferreira:
So, there are 8 tests failing, all of them show:

```
 // if this fails, it means the issue's been fixed
```

I believe those are issues that happen only on native tests...
I just added a test to skip that part for Node and all tests were working.

sidferreira:
@radex note: I rebased and force pushed!

radex:
Thanks @sidferreira for the amazing work! This will be really useful for building tools… heck, maybe someone's crazy enough to use WatermelonDB as a Node server-side DB (I wouldn't recommend it, but would be cool, no?!)

radex:
@sidferreira BTW: I want to send you some 🍉 stickers - if you want 'em, email me with your address :)

PR #732, opened 2020-06-18
Q.on nesting

radex:


src/QueryDescription/test.js
```diff
@@ -40,48 +37,13 @@ describe('QueryDescription', () => {
...
-          comparison: {
-            operator: 'eq',
-            right: { value: null },
-          },
-        },
+        { type: 'where', left: 'col1', comparison: { operator: 'eq', right: { value: 'val1' } } },
+        { type: 'where', left: 'col2', comparison: { operator: 'eq', right: { value: 2 } } },
+        { type: 'where', left: 'col3', comparison: { operator: 'eq', right: { value: true } } },
+        { type: 'where', left: 'col4', comparison: { operator: 'eq', right: { value: false } } },
+        { type: 'where', left: 'col5', comparison: { operator: 'eq', right: { value: null } } },
```
radex, line 44: reformatting, nothing to see

src/QueryDescription/test.js
```diff
@@ -302,35 +177,25 @@ describe('QueryDescription', () => {
     ])
```
radex, line 177: reformatting, nothing to see

src/QueryDescription/test.js
radex: nothing to see in this file until a comment that says there's stuff to see :D

src/QueryDescription/test.js
```diff
@@ -302,35 +177,25 @@ describe('QueryDescription', () => {
...
     expect(query).toEqual({
       where: [
```
radex, line 179: ok, start looking here

src/adapters/sqlite/encodeQuery/test.js
```diff
@@ -30,7 +31,12 @@ describe('SQLite encodeQuery', () => {
       Q.where('col5', null),
     ])
     expect(encodeQuery(query)).toBe(
-      `select "tasks".* from "tasks" where "tasks"."col1" is 'value "''with''" quotes' and "tasks"."col2" is 2 and "tasks"."col3" is 1 and "tasks"."col4" is 0 and "tasks"."col5" is null and "tasks"."_status" is not 'deleted'`,
+      `select "tasks".* from "tasks" where "tasks"."col1" is 'value "''with''" quotes'` +
```
radex, line 34: it's all reformatting until the comment that says it's not

src/adapters/sqlite/encodeQuery/test.js
```diff
@@ -99,16 +132,32 @@ describe('SQLite encodeQuery', () => {
...
+        ` join "projects" on "projects"."id" = "tasks"."project_id" where "projects"."left_column" <= "projects"."right_column"` +
+        ` and ("projects"."left2" > "projects"."right2"` +
+        ` or ("projects"."left2" is not null` +
+        ` and "projects"."right2" is null))` +
+        ` and "projects"."_status" is not 'deleted'` +
+        ` and "tasks"."_status" is not 'deleted'`,
     )
   })
-  it('encodes like and notLike queries', () => {
+  it(`encodes on nested in and/or`, () => {
```
radex, line 144: ok, real changes start here

radex:
Yet another big PR. Sorry about it. this shit's so interlocked, I don't think I could have unwound it enough to make it multiple steps, that each actually works.

So i'll help by explaining the big picture:

1. before, Q.ons were a special kind of clause. they were only allowed in the top level of a query, and were always "AND"ed with other top-level conditions
2. having Q.ons at the top simplified things because it was trivial to check if there are any Q.ons at all in the query. Right now, it would require recursively searching for it, which I wanted to avoid (for simplicity of implementation and performance). So I require explicit Q.experimentalJoinTables(xxxx) if you nest Q.ons (if you don't, it's not required - for backwards compatibility). Well, I still need to recursively search for Q.ons to implement queryWithoutDeleted, so maybe I'll remove this restriction later. (BTW: If this restriction is broken, encodeMatcher/encodeQuery will fail once they get to Q.on)
3. queryWithoutDeleted is now really complicated, because I need to insert `_status != deleted` on all join tables, but I can't do it in top scope, it has to be where Q.on resides (otherwise something like `tasks.query(or(is_followed=true, on(tag_assignments, tag_id=x)))` would not work right, since the task would HAVE to have a tag assignment that's not deleted, ignoring the other ORed condition. I don't like it, but don't have a better idea now than to recursively search for Q.ons and rewrite them. Maybe we'll move to filtering out deleted records on adapter/query encoding layer, instead of on Query layer. I don't know yet.
3. SQLite implementation was fairly straightforward, although I found a potential problem which I explained in code comments
4. LokiJS implementation required significant reworking because again, we start to encode a Loki query, but to actually execute the sub-queries on joined tables, we have to recursively search for those tables. But I got it to the point where adding nested Q.ons (e.g. query on a grandparent) will be really simple (i think)

radex:
don't mind the red CI - it's broken on master, hopefully will fix on JSI PR

src/QueryDescription/test.js
kokusGr: Thanks, that was really helpful :)

src/QueryDescription/index.js
```diff
@@ -340,13 +365,51 @@ export function buildQueryDescription(clauses: Clause[]): QueryDescription {
...
+  const conditions: Where[] = wheres.map(queryWithoutDeletedImpl)
+  // $FlowFixMe
+  const ons: On[] = conditions.filter(clause => clause.type === 'on')
+  const onTables = uniq(ons.map(clause => clause.table))
+  return conditions.concat(onTables.map(whereNotDeletedJoin))
+}
+
+function conditionWithoutDeleted(clause: Where): Where {
+  if (clause.type === 'on') {
+    const onClause: On = clause
```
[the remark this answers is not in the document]
radex: Yes, it is, it's not smart enough in this context to understand that if clause.type === 'on', then clause is the `On` of `Where = WhereDescription | And | Or | On` alternative

src/QueryDescription/index.js
```diff
@@ -340,13 +365,51 @@ export function buildQueryDescription(clauses: Clause[]): QueryDescription {
...
+    }
+  }
+  return queryWithoutDeletedImpl(clause)
+}
+
+function queryWithoutDeletedImpl(clause: Where): Where {
+  if (clause.type === 'and') {
+    return {
+      type: 'and',
+      conditions: conditionsAndJoinTablesWithoutDeleted(clause.conditions),
```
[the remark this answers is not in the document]
radex: right now, if you have multiple `on` conditions, you'll write it like this:

```
on('projects', foo=bar),
on('projects', bar=baz),
on('projects', baz=blah),
```

so if we treated `and` the same way as `or`, we'd get:

```
and(on('projects', foo=bar), on('projects', _status != 'deleted'))
and(on('projects', bar=baz), on('projects', _status != 'deleted'))
and(on('projects', baz=blah), on('projects', _status != 'deleted'))
```

and since we know we're in an `and`, we can skip the duplicate conditions. They probably wouldn't be harmful for SQLite, but probably would be harmful for performance for LokiJS.

I think the real solution is to introduce this syntax:

```
on('projects', [
  foo = bar,
  bar = baz,
  baz = blah,
])
```

then we wouldn't need this and a lot of other special cases, in addition to making yet other cases more performant on LokiJS :) So this is what I will do - and remove the special cases

src/adapters/lokijs/worker/performJoins/index.js
```diff
@@ -0,0 +1,35 @@
...
+import type { DirtyRaw } from '../../../../RawRecord'
+
+type QueryPerformer = (join: LokiJoin) => DirtyRaw[]
+
+function performJoinsImpl(query: LokiRawQuery, performer: QueryPerformer): LokiRawQuery {
+  if (!query) {
+    return query
+  } else if (query.$join) {
+    const join: LokiJoin = query.$join
+    const records = performer(join)
```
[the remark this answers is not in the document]
radex: yep, this is to make this a pure function, so that it's easy to test this logic without mocking or testing at the whole database level. Alternatively, we'd still need to pass `loki: Loki` as parameter, so there's no difference in the end, and this one's cleaner

src/QueryDescription/index.js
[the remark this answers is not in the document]
kokusGr: Thanks for explanation :) So this is optimisation, not that `and` needs to be treated differently? But isn't `or` similar? For:
```
Q.or(
  Q.on("projects", foo=bar),
  Q.on("projects", bar=baz),
  Q.on("projects, baz=blah),
)
```
we get:
```
Q.or(
  Q.and(Q.on("projects", foo=bar), Q.on("projects", _status != "deleted")),
  Q.and(Q.on("projects", bar=baz), Q.on("projects", _status != "deleted")),
  Q.and(Q.on("projects", baz=blah), Q.on("projects", _status != "deleted")),
)
```
right? So ideally we would want sth like this:
```
Q.and(
  Q.or(
    Q.on("projects", foo=bar),
    Q.on("projects", foo=bar),
    Q.on("projects", foo=bar)
  ),
  Q.on("projects", _status != "deleted")
)
```
Or this won't work? New syntax looks nice :)

src/adapters/lokijs/worker/performJoins/index.js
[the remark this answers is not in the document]
kokusGr: Ok, thanks :)

kokusGr:
@radex Merging, but please take a look at my last comment :)

src/QueryDescription/index.js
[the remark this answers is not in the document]
radex: this would work only if all elements of `Q.or` are `Q.on`s on the same table. So we get even more edge cases to cover. Bah, better to fix the root cause

PR #736, opened 2020-06-24
Query unsafe SQL & Loki expressions

radex:


radex:
> - [Query] `Q.unsafeSqlExpr()` and `Q.unsafeLokiExpr()` are introduced to allow adding bits of queries
>      that are not supported by the WatermelonDB query language without having to use `unsafeFetchRecordsWithSQL()`.
>      See docs for more details

Known limitation: observation will throw an error for some queries that use this. I'll fix it, but not in this PR.

I don't recall a specific situation where we'd need it in NT, but I bet that there absolutely are use cases that we (less efficiently and less nicely) solve at a higher level (rx/js filtering), but that we could do both in SQL and in Loki (but it's weird/unusual enough that it's not worth developing a proper Q. function, testing it, etc... -- or maybe as a temporary solution from the app dev team until I implement the proper solution). And I'm sure many open source users are gonna love it.

I think it's generally useful to have these finer-grained "escape hatches" so that WatermelonDB doesn't constrain us. So I'll also introduce something similar (raw SQL) in schema/migration definitions -- I don't yet know exactly what's our use case, but I think once we have the option, we're gonna use it. (e.g. to fine-tune some SQLite options that aren't necessarily worth including in a high level API)

src/QueryDescription/index.js
```diff
@@ -247,22 +250,32 @@ export function where(left: ColumnName, valueOrComparison: Value | Comparison):
...
-const acceptableClauses = ['where', 'and', 'or', 'on']
-const isAcceptableClause = (clause: Where) => acceptableClauses.includes(clause.type)
+export function unsafeSqlExpr(sql: string): SqlExpr {
+  invariant(typeof sql === 'string', 'Value passed to Q.unsafeSqlExpr is not a string')
+  return { type: 'sql', expr: sql }
+}
 
-export function and(...clauses: Where[]): And {
+export function unsafeLokiExpr(expr: any): LokiExpr {
+  invariant(expr && typeof expr === 'object', 'Value passed to Q.unsafeLokiExpr is not an object')
```
kokusGr, line 259: This also allows to pass an Array. Is it ok?

kokusGr:
>I think it's generally useful to have these finer-grained "escape hatches" so that WatermelonDB doesn't constrain us.

100% agree :)

src/QueryDescription/index.js
radex: fixed

radex:
> if I would run it web I would actually get an error because of unsafeSqlExpr? In that case wouldn't it be better to ignore sql clause on web and loki clause on native? Or am I reading it wrong? :)

maybe it would be better to have this automatic behavior. not sure. for now, let's be explicit -- I changed docs to be less confusing

kokusGr:
Ok, we will see if it becomes an issue. Merging, thanks :)

PR #743, opened 2020-06-29
LokiJS column comparisons v2

radex:


radex:
LokiJS didn't use to have any syntax for column comparison queries, so it was always implemented as a big hack - we skipped it when querying the DB, and then filtered again using a matcher function in JS world. This worked, but was unperformant, and broken when restrictions on Q.on were lifted.

Well. Instead of trying to fix my awful unperformant hack, I fixed the problem at the root: https://github.com/techfort/LokiJS/pull/847 . Was much easier than expected (took me 2 hours), I wish I did it in 2018 and never implemented any hacks! And look how much cleaner the code is! Okay, there are some weird edge cases that complicate the query, but still!

PR #746, opened 2020-07-02
Unsafe SQL schema, migration

radex:
You can now pass `unsafeSql` parameters to schema builder and migration steps to modify SQL generated to set up the database or perform migrations. There's also new `unsafeExecuteSql` migration step. 

This is essentially an escape hatch so that if you really need it, you can do it without modifying WatermelonDB or introducing one-off 🍉 syntax (with all the bloat this entails) just for some rare case.

Example use cases:

- custom SQLite data types (json, geo, etc)
- optimizations (indices on multiple columns, `without rowid`)
- compatibility with older systems


docs-master/Schema.md
```diff
@@ -88,6 +88,30 @@ However, if you rarely query all comments by its author, indexing `author_id` is
 
 In general, most `_id` fields are indexed. Sometimes, `boolean` fields are worth indexing if you often use it for queries. However, you should almost never index date (`_at`) columns or `string` columns.
 
+## Advanced
+
+### Unsafe SQL schema
+
+If you want to modify the SQL used to set up the SQLite database, you can pass `unsafeSql` parameter
+to `tableSchema` and `appSchema`. This parameter is a function that receives SQL generated by Watermelon,
```
kokusGr, line 96: I think this change indirectly makes SQL generated by Watermelon kinda a public API, since now apps can search and replace specific part etc. What will be our approach to changes in generated SQL? Will these be marked as breaking changes? If not maybe we should add this info to documentation to avoid surprises.

radex: I see your point, but I don't want to overcomplicate this. IMO `unsafe` means unsafe, can break at any moment. But I agree it's worth noting that in the docs, and we can make it a habit to note encode schema changes in changelog (but not as breaking changes in the semver sense)

kokusGr: Yeap, seems like a reasonable approach :)

Issue #608, opened 2020-01-19
React-native example app not working


sofiageo:
Hi, I wanted to try the react-native example for Android but it's not working at the moment. 

First, WatermelonDB version is at 0.10.1 - which means that is still affected by #276 
But even correcting this it seems that the metro server can't communicate with my device. I guess maybe updating some libraries / dependencies can fix this? I see a PR #513 but no activity there. I'm not sure what else I might be missing. I can probably help testing the example if the PR is finalized / accepted.

AugustoAleGon:
@sofiageo try to use this. It is working just on iOS. You can enable for Android if you want. https://github.com/AugustoAleGon/watermelonDBSample

sofiageo:
@AugustoAleGon 

Thanks, it took some time but I manage to get it working.

AugustoAleGon:
@sofiageo I am glad that helps! If you manage to work on Android just send a pr for it. 💪

sofiageo:
> @sofiageo I am glad that helps! If you manage to work on Android just send a pr for it. muscle

I'm having some ISP issues at the moment at home so I can send a PR after they are resolved. However, I also had to upgrade react to 16.9.0 + react native to 0.61.5 while I was trying to get it working. If it's OK with you I will do it as soon as possible (when I have proper access to internet again). I will open an issue on your repository so we don't spam here :)

Of course I'm still interested in setting up the official watermelonDB example :)

mjstelly:
The example RN is way out of date. @sofiageo said he's working on updating it, but that was a while ago. If I had the time, I would. I'm too busy researching offline storage solutions.

sofiageo:
Hi @mjstelly - Check the demo from @AugustoAleGon - I had sent the PR in January and I guess it should still work. I was never sure about the status of the code here so I never tried to fix anything. However I've managed to use WatermelonDB for a project and all went well at the end (apart from fast deletes not working yet). I would recommend it for anyone lurking here :)

radex:
@sofiageo hey, would you be interested in updating the built-in react native example to the latest versions? that would be very helpful to a lot of people using 🍉 :)

mjstelly:
I have an RN project already started. I’ll post it to my public github. You can fork it if you wish. @radex @sofiageo

sofiageo:
[...]

PR #751, opened 2020-07-06
Upgraded react-native and dependencies versions. #608

sofiageo:
This took a bit longer than expected. It's running at the moment for Android but some things can still be improved. (For example navigation library). I have not tested iOS yet.


radex:
@sofiageo Excellent work, thank you! ❤️

PR #756, opened 2020-07-09
Add cassert dependency to Sqlite.cpp

jakebloom:
After updating to the latest version of watermelondb, our builds started failing with `Use of undeclared identifier 'assert'` errors coming back to this file. 

We are pretty light on native code, so we hadn't included it elsewhere in our project, we should include the dependency explicitly here.

After adding this line, the build worked properly and our app ran as expected

### DIFF
--- a/native/shared/Sqlite.cpp
+++ b/native/shared/Sqlite.cpp
@@ -1,5 +1,6 @@
 #include "Sqlite.h"
 #include "DatabasePlatform.h"
+#include <cassert>
 
 namespace watermelondb {
 


jakebloom:
ping @radex

radex:
@jakebloom Hey! I seem to recall seeing `cassert`  issues once, but I don't remember when, why, and why the issue went away. I assume you're talking about iOS? What's the Watermelon, RN version? C++ toolchain (C++14?)

jakebloom:
Yep, iOS - React Native 0.61, C++14, latest WatermelonDB release. This is using the iOS 14 developer beta though, so there might have been changes made there

radex:
okay then!

PR #777, opened 2020-07-30
Performance 07/30

radex:


radex:
@rozPierog CR pls. The changes here are really basic - really just passing things through to LokiJS where the real magic happens.

PR #781, opened 2020-08-06
Selective sync 2

radex:
Adds the ability to override conflict resolution… we'll need that to implement Selective Sync v2 in NT, but it has other uses we (and others) might benefit from in the future..


radex:
@rozPierog mind taking a look?

radex:
@kokusGr CR pls?

kokusGr (approved):
LGTM 👍

PR #791, opened 2020-08-17
Performance improvements & fixes

radex:


babel.config.js
```diff
@@ -1,5 +1,23 @@
 const plugins = [
-  '@babel/plugin-transform-modules-commonjs',
+  [
+    '@babel/plugin-transform-runtime',
+    {
+      helpers: true,
+      // regenerator: true,
```
kokusGr, line 6: I think that `true` is the default value.

rn-cli.config.js
```diff
@@ -1,11 +1,18 @@
...
+const glob = require('glob-to-regexp')
 
 const getBlacklistRE = () => {
-  // delete __tests__ from the default blacklist
-  const defaultPattern = blacklist()
+  // ignore dist/, dev/
+  const defaultPattern = blacklist([
+    glob(`${path.resolve(__dirname, '..')}/dist/*`),
+    glob(`${path.resolve(__dirname, '..')}/dev/*`),
+    glob(`${path.resolve(__dirname, '..')}/example/*`),
```
kokusGr, line 10: Hmm I am not sure if this is what you want but to ignore all nested files I think you should use
```
glob(`${path.resolve(__dirname, '..')}/example/**`, { globstars: true })
```

babel.config.js
```diff
@@ -52,9 +76,6 @@ module.exports = {
     production: {
       plugins: [
         ...plugins,
-        // console.log is expensive for performance on native
-        // we don't want it on web either, but it's useful for development
-        ['transform-remove-console', { exclude: ['error', 'warn'] }],
```
kokusGr, line 57: Why we no longer want that?

kokusGr (approved):
LGTM 👍 Just have some questions and ready to merge.

babel.config.js
radex: I don't think it actually did anything, AFAIR all console.logs showed up anyway. And anyway, all logging goes through `logger`, and not directly to console.log, so applications can choose whether or not to silence logs

babel.config.js
radex: noted. i don't want to change it now, because I'd have to re-test everything, but i'll pay closer attention next time i'm looking at this

rn-cli.config.js
radex: my bad. it seems that this fix is still broken :( … and it's not because of this, because Metro trips up on package.json file which is directly in /dev/, for example. But you're right, i'll change it later.

kokusGr:
Thanks for your answers :) Merging.

PR #801, opened 2020-08-29
Allowing LokiJS autosave to be configured

jrounsav:
The LokiJS autosave feature is initiating an infinite loop during jest tests when running `jest.runAllTimers()`.

Allowing it to be configured so that it can be disabled when mocking out the adapter.


radex:
@jrounsav hey, this looks good, but please modify `type LokiAdapterOptions = ` signature to include the `autosave` option and add a mention about this to Changelog-unreleased.md

jrounsav:
@radex Thank you. Signature and changelog updated 👍

radex:
Nice! If I were you, I'd instead add an option to pass an extra object with properties to append at the end of Loki constructor -- would be more powerful... But oh well, if someone needs it, they'll add it :)

Issue #799, opened 2020-08-27
[Question] [iOS] Sharing database across targets


thiagobrez:
Hi!

Is sharing the database across Xcode targets something possible? I've created a Share Extension target and wanted to use the same db as my main target. I've sucessfully set up Watermelon, but when querying for records the collections exist but are empty. Maybe I'm missing something? Already tried exporting the Database and Adapter instances as singletons, with no luck. It's worth noting that I'm using two separate DatabaseProviders.

radex:
you need to set up db path to be within a shared container.

thiagobrez:
Ok! Appreciate your quick response. But is this something I'll have to write in Watermelon files or the Extension files? Looking into it, my targets are already in the same App Group and setting the shared container seems fairly easy, but I'm kind of lost where to start. Could you point me in the right direction?

Anyway, thanks for your help!

radex:
@thiagobrez I'll try my best to help you with this, if you agree to write it up nicely in https://github.com/Nozbe/WatermelonDB/tree/master/docs for others to see

thiagobrez:
@radex For sure! I'd love to contribute

radex:
Okay:

1. You need to get an URL for the app group:

```swift
FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "group.MY-GROUP_ID").appendingPathComponent("MY-APP.db")
```

(you need to generate an app group id via apple developer site; in Obj-C, start typing NSFileManager, and follow autocompletions and you'll figure it out)

2. You need to pass this URL to React Native land via a Native Module:

```swift
@objc(constantsToExport)
    var constantsToExport: NSDictionary {
        return [
            "databasePath": databaseUrl.path,
        ]
    }
```

see this guide for making a native module + ObjC equivalent: https://reactnative.dev/docs/native-modules-ios#exporting-constants

3. when initializing DB from JS, use:

```js
  return new SQLiteAdapter({
    // Use db in app group folder on iOS, standard folder on Android
    dbName: platform.isiOS ? NativeModules.MyAppPlugin.databasePath : 'nozbe',
    // ... rest of parameters
  })
```

looking forward to that mini-guide more nicely explained than this!

thiagobrez:
Perfect! I'll start digging into this in a few hours and once I get it working will open a PR into the docs so we can review and discuss.

diegolmello:
Hey! :)
[...]

PR #809, opened 2020-09-08
iOS - Sharing db across targets

thiagobrez:
This PR adds a tutorial on how to share Watermelon across iOS targets, as discussed in #799.

All props goes to @radex, @diegolmello and @djorkaeffalexandre !


radex:
Amazing, thank you! ❤️

PR #816, opened 2020-09-14
DatabaseProvider: Remove cycle dependency warning

gmonte:
I created a specific file to create and provide the DatabaseContext, resolving the issue.
All tests was ok here.


src/DatabaseProvider/index.js
```diff
@@ -1,9 +1,7 @@
 // @flow
 import React from 'react'
 import Database from '../Database'
-
-export const DatabaseContext = (React.createContext(): any)
-const { Provider, Consumer } = DatabaseContext
+import { Provider } from './DatabaseContext'
```
radex, line 4: can you export DatabaseContext from here? Otherwise, you're making an API-breaking change, in case someone is using this export

radex:
Looks good! If you'd like, add a note to Changelog-Unreleased.md about the fix with your GH nick

radex:
Thanks!

PR #828, opened 2020-09-23
[FIX] Get date as instance of Date

djorkaeffalexandre:
When a date is cached, it's returning as a string instead of a Date instance.


radex:
@djorkaeffalexandre ok, can you add a note to Changelog-Unreleased? thanks

djorkaeffalexandre:
@radex, done! Thanks :)

PR #835, opened 2020-10-06
fix type of OnFunction to accept Q.and and Q.or

mlecoq:
I need to use `Q.and` in Q.on 

```
 Q.on( 'projectDocument_folders', Q.and(Q.where('folder_type', Q.eq('plan')), Q.where('project_id', Q.eq(project.id))),),
```

Only `Q.where` is accepted according to defined types


radex:
thanks!

Issue #841, opened 2020-10-12
incorrect create/update tracking 


leo-tavares:
I follow the [doc](https://nozbe.github.io/WatermelonDB/Advanced/CreateUpdateTracking.html) about created and updated but, when watermelonDB create da instance he set an created_at with ``` "1970-01-01T00:00:00.000Z" ``` . 
I checked the date of my computer and the emulator of these devices are correct.  Where am I going wrong?

### snippet

```typescript
      await database.action(async () => {
        const meal = await mealCollection.create((m) => {
          m.name = 'dinner';
        });
        await meal.subAction(() =>
          meal.addMealItems([{name: 'chicken'}, {name: 'rice'}]),
        );
      });

```

### database related files (models, schema, initialization) 

[link](https://github.com/leo-tavares/log-alimentar-mobile/tree/master/src/database)

#### link of project
[log-alimentar](https://github.com/leo-tavares/log-alimentar-mobile)

leo-tavares:
The problem was in the definition of my model. apparently we cannot put the same name for these special fields

### before (wrong way)
```typescript
   @readonly
   @date ('created_at')
   created_at !: Date;

   @readonly
   @date ('updated_at')
   updated_at !: Date;
```
### after (correct way)
```typescript
  @readonly
  @date('created_at')
  createdAt!: Date;

  @readonly
  @date('updated_at')
  updatedAt!: Date;
```

PR #842, opened 2020-10-12
Update CreateUpdateTracking.md

leo-tavares:
I added a note explaining the importance of not using the same name for the 'created_at' and 'updated_at' columns.
I made this small mistake and it took me a few hours to solve the problem.
I believe that with this note, it will help someone

### DIFF
--- a/docs-master/Advanced/CreateUpdateTracking.md
+++ b/docs-master/Advanced/CreateUpdateTracking.md
@@ -15,7 +15,7 @@ You can add per-table support for create/update tracking. When you do this, the 
 
 **Notes**:
  - you _don't have to_ enable both create and update tracking. You can do either, both, or none.
- - In your model, you cannot name the columns created_at and updated_at with a 'created_at' and 'updated_at' respectively. this causes problem with date creation and update.
+ - In your model, these fields need to be called createdAt and updatedAt respectively.
 
 ### How to do this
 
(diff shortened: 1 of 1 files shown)


docs-master/Advanced/CreateUpdateTracking.md
```diff
@@ -13,7 +13,9 @@ You can add per-table support for create/update tracking. When you do this, the
 
 - When you display to the user when a thing (e.g. a Post) was modified
 
-**Note**: you _don't have to_ enable both create and update tracking. You can do either, both, or none.
+**Notes**:
+ - you _don't have to_ enable both create and update tracking. You can do either, both, or none.
+ - In your model, you cannot name the columns created_at and updated_at with a 'created_at' and 'updated_at' respectively. this causes problem with date creation and update.
```
radex, line 18: I read this twice, and honestly I'm really confused what you mean here

radex (commented):
Thank you for your contribution, but please explain what you really mean, because despite me being the person who has implemented create/update tracking, I feel more confused instead of more enlightened after reading it

docs-master/Advanced/CreateUpdateTracking.md
leo-tavares: Hello @radex !

A few days ago ... I faced a bug. The time of created_at and updated_at was always "1970-01-01T00: 00: 00.000Z". I spent hours trying to understand the problem until I decided to modify my model, especially 'created_at' and 'updated_at' I called them createdAt and updatedAt and the bug will resolved. 

I open an issue, look => https://github.com/Nozbe/WatermelonDB/issues/841 .

my English is not good. But i want help community :D

radex: @leo-tavares I see! I think it would be much clearer if you phrased it as something like that the defined fields must be called precisely createdAt updatedAt

radex:
cool, thanks for your contribution!

PR #860, opened 2020-11-09
Return record in update method on Model.

pranavbabu:
This PR changes behavior of `update` to be consistent with the `Collection.create` method where it returns the record. I did this because it would be useful to get back record after update, to further manipulations like invoking subactions etc...


radex:
that's reasonable! here's more work that I need before this is mergeable:

- [ ] Flow/TS type signature
- [ ] A test

radex:
Looks good! Thanks @pranavbabu !

PR #871, opened 2020-11-26
Diagnostics, etc.

radex:


radex:
- [LokiJS] `useWebWorker` and `useIncrementalIndexedDB` options are now required (previously, skipping them would only trigger a warning)
- [adapters] `onSetUpError: Error => void` option is added to both `SQLiteAdapter` and `LokiJSAdapter`. Supply this option to catch initialization errors and offer the user to reload or log out
- [SQLiteAdapter] `synchronous:true` option is now deprecated and will be replaced with `experimentalUseJSI: true` in the future. Please test if your app compiles and works well with `experimentalUseJSI: true`, and if not - file an issue!
- [Resilience] Added extra diagnostics for when you encounter the `Record ID aa#bb was sent over the bridge, but it's not cached` error and a recovery path (LokiJSAdapter-only). Please file an issue if you encounter this issue!
- Added an experimental mode where a broken database state is detected, further mutations prevented, and the user notified
- experimental `@_reader`, `@_writer` decorators

kokusGr (approved):
LGTM 👍

kokusGr:
Shiet sorry didn't notice it's dependant branch :/

PR #881, opened 2020-12-03
android: build.gradle use safeExtGet boilerplate

zxcpoiu:
Android build.gradle

#### Use `safeExtGet` boilerplate

To consist with most libs syntax

(my bad, we still need user to specify kotlin version in their build.gradle)

### DIFF
--- a/native/android/build.gradle
+++ b/native/android/build.gradle
@@ -6,13 +6,17 @@ def DEFAULT_BUILD_TOOLS_VERSION = "28.0.3"
 def DEFAULT_MIN_SDK_VERSION = 16
 def DEFAULT_TARGET_SDK_VERSION = 28
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 android {
-    compileSdkVersion rootProject.hasProperty('compileSdkVersion') ? rootProject.compileSdkVersion : DEFAULT_COMPILE_SDK_VERSION
-    buildToolsVersion rootProject.hasProperty('buildToolsVersion') ? rootProject.buildToolsVersion : DEFAULT_BUILD_TOOLS_VERSION
+    compileSdkVersion safeExtGet("compileSdkVersion", DEFAULT_COMPILE_SDK_VERSION)
+    buildToolsVersion safeExtGet("buildToolsVersion", DEFAULT_BUILD_TOOLS_VERSION)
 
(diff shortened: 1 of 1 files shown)


radex:
@rozPierog pls cr & merge if good

radex:
@rozPierog ⬆️

native/android/build.gradle
```diff
@@ -6,13 +6,17 @@ def DEFAULT_BUILD_TOOLS_VERSION = "28.0.3"
 def DEFAULT_MIN_SDK_VERSION = 16
 def DEFAULT_TARGET_SDK_VERSION = 28
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
```
rozPierog, line 10: Wouldn't it require current users of WatermelonDB to move from 
```groovy

project.ext {
    compileSdkVersion = 28
    buildToolsVersion = '28.0.3'
    targetSdkVersion = 26
    minSdkVersion = 23
}
```
assigment to 
```groovy
buildscript {
  ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 23
        compileSdkVersion = 28
        targetSdkVersion = 28
    }
[...]
}
```
?

rozPierog (changes requested):

native/android/build.gradle
zxcpoiu: From my understanding, this should not make any differences to the sub modules.

`project.ext.has("foo")` is [equivalent](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtraPropertiesExtension.html#org.gradle.api.plugins.ExtraPropertiesExtension:has(java.lang.String)) to `project.hasProperty("foo")`

And in the rootProject's `build.gradle`, the following definitions all can be accessed in sub module via `safeExtGet`

```

// --- top level
project.ext.customExtA = "YES"

// --- top level ext block
ext {
    customExtB = "YES"
    ...
}

// --- define ext inside buildscript block
buildscript {
    ext.customExtC = "YES"
}
```

It's about evaluation order at all. (The buildscript {} block is evaluated before anything else in the script)

Correct me if I missed something :)

radex:
@zxcpoiu Can you merge master into this branch so we can see if CI passes?

zxcpoiu:
Done! But maybe should restart the iOS tests

zxcpoiu:
Closed in favor of #977

PR #883, opened 2020-12-04
Fix database#batch(records) types

keima:
In [Batch Updates section of WatermelonDB documentation](https://nozbe.github.io/WatermelonDB/Actions.html#batch-updates),

> You can pass falsy values (null, undefined, false) to batch — they will simply be ignored.

But below code written in TypeScript causes type error on latest version:

```
database.batch(user, null, false, undefined)
```

> TS2345: Argument of type '[User, null, false, undefined]' is not assignable to parameter of type 'Model[] | null[] | void[] | false[] | Promise<void>[]'.   Type '[User, null, false, undefined]' is not assignable to type 'Promise<void>[]'.     Type 'false | User | null | undefined' is not assignable to type 'Promise<void>'.       Type 'undefined' is not assignable to type 'Promise<void>'.

(note: `User` is database model)

This patch fixes above error.

### DIFF
--- a/CHANGELOG-Unreleased.md
+++ b/CHANGELOG-Unreleased.md
@@ -14,5 +14,6 @@
 ### Fixes
 
 - [Typescript] Fixed type on OnFunction to accept `and` in join 
+- [Typescript] Fixed type `database#batch(records)`'s argument `records` to accept mixed types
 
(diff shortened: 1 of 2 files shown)


radex:
thanks!

PR #805, opened 2020-09-03
Add android autolinking config on react-native.config.js

keima:
On Android, Autolining is not working properly Due to android config is null.

```
~/g/RNSandbox63 ❯❯❯ yarn --silent react-native config | jq '.dependencies."@nozbe/watermelondb"'
{
  "root": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb",
  "name": "@nozbe/watermelondb",
  "platforms": {
    "ios": {
      "sourceDir": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/native/ios",
      "folder": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb",
      "pbxprojPath": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/native/ios/WatermelonDB.xcodeproj/project.pbxproj",
      "podfile": null,
      "podspecPath": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/WatermelonDB.podspec",
      "projectPath": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/native/ios/WatermelonDB.xcodeproj",
      "projectName": "WatermelonDB.xcodeproj",
      "libraryFolder": "Libraries",
      "sharedLibraries": [],
      "plist": [],
      "scriptPhases": []
    },
    "android": null
  },
  "assets": [],
  "hooks": {},
  "params": []
}

```

So, adding Android config to working properly.

```
~/g/RNSandbox63 ❯❯❯ yarn --silent react-native config | jq '.dependencies."@nozbe/watermelondb"'
{
  "root": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb",
  "name": "@nozbe/watermelondb",
  "platforms": {
    "ios": {
      "sourceDir": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/native/ios",
      "folder": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb",
      "pbxprojPath": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/native/ios/WatermelonDB.xcodeproj/project.pbxproj",
      "podfile": null,
      "podspecPath": "/Users/keima/git/RNSandbox63/node_modules/@nozbe/watermelondb/WatermelonDB.podspec",
[...]

PR #884, opened 2020-12-07
Add react-native.config.js to npm packages

keima:
Previously, I implemented Autolinking on Android.

- https://github.com/Nozbe/WatermelonDB/pull/805

But I forgot to packaging `react-native.config.js`.
So v0.20 is NOT supported Autolinking on Android.

This patch fixes this problem (may be... I not yet familiar with Watermelon build system.).


radex:
oops, you're right, I didn't notice this. I think this time it should work :)

PR #898, opened 2020-12-17
🏷 Add Missing Props for SQLiteAdapterOptions

castroCrea:
TS error with web installation 
```
// First, create the adapter to the underlying database:
const adapter = new SQLiteAdapter({
  schema,
  // dbName: 'myapp', // optional database name or file system path
  // migrations, // optional migrations
  synchronous: true, // synchronous mode only works on iOS. improves performance and reduces glitches in most cases, but also has some downsides - test with and without it
  // experimentalUseJSI: true, // experimental JSI mode, use only if you're brave
})
```


radex:
thanks!

Issue #931, opened 2021-01-25
unsafeFetchRecordsWithSQL & fetch - possible race condition?


jspizziri:
Alright, I've spent a fairly reasonable amount of time attempting to debug this including putting log statements in the `node_modules` source of `watermelondb`. 

In a nutshell the _symptom_ of the issue is that I'm getting the classic `[Diagnostic error: Record ID foos#12345 was sent over the bridge, but it's not cached` error message. It feels as though there might be some sort of race condition going on based on what I'm seeing, but I'll let the experts be the judge of that. If I can provide any more information I'd be happy to. Here's the breakdown:

I'm using `unsafeFetchRecordsWithSQL` to fetch a particular record, and separately fetching a larger set of that records type with a traditional query/fetch. These two operations happen asynchronously in the context of one another. The minimum reproducible code that I've created is something to the effect of the following (please note: in reality the unsafe query is more complex, but the below appears to do the job of reproducing):

```ts
collection.unsafeFetchRecordsWithSQL(`
        SELECT f.*
        FROM foos AS f
        WHERE f.id = '12345'
        LIMIT 1
    `)
    .catch((err) => {
        console.error('Error when executing unsafe fetch', err);
    })
;

// `12345` will also be fetched by this query
collection.query()
    .fetch()
    .catch((err) => {
        console.error('Error when executing safe collection fetch', err)
    })
;
```

In this case, the _safe_ fetch will throw the `Diagnostic error: Record ID foos#12345 was sent over the bridge, but it's not cached]` error.

When logging, it appears that while the record cache has an ID in the map for `12345` it points to `undefined` which I can only assume is the result of the fact that the `unsafeFetchRecordsWithSQL` has begun, but hasn't fully resolved.

Thoughts? Thanks in advance!

radex:
interesting, this definitely sounds like a bug. I understand that the unsafe fetch _succeeds_ eventually? (If it fails, then this is a problem, Watermelon is not resilient enough to recover from illegal queries

jspizziri:
@radex correct, the unsafe query does succeed eventually, but the other does not. If you want to point me in the right direction I'd be happy to do some more sleuthing on this, I know you've got a lot on your plate maintaining this whole thing.

radex:
Yes I do, and so `fixing your own problems` is very much appreciated, especially if it comes with a pull request :) But happy to assist, I don't want any observable race conditions in this project :)

[...]

PR #932, opened 2021-01-25
fix(unsafefetchrecordswithsql): fix an issue with unsafeFetchRecordsW…

jspizziri:
…ithSQL causing a race-condition

This commit fixes an issue where `unsafeFetchRecordsWithSQL` when used in conjunction with a
standard `fetch` in an asynchronous manner while fetching sets of intersecting records  was
resulting in an error due to a race-condition.

https://github.com/Nozbe/WatermelonDB/issues/931

### DIFF
--- a/src/Collection/index.js
+++ b/src/Collection/index.js
@@ -1,5 +1,5 @@
 // @flow
-
+import type { SQLDatabaseAdapter } from '../adapters/type'
 import { Observable, Subject } from '../utils/rx'
 import invariant from '../utils/common/invariant'
 import noop from '../utils/fp/noop'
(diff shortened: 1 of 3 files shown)


jspizziri:
@radex ,

Here's a rough PR. I've confirmed within my project that this does fix the issue I was experiencing as you expected. Please provide some feedback on the implementation.

Lastly, I could use a touch of help with writing the appropriate test. As it currently stands, the test that I wrote doesn't actually work. I'm not sure what functions need to be mocked appropriately to replicate the issue.

jspizziri:
@radex any chance you could take a look at this?

radex:
Looks good, thanks for doing this! but CI fails -- run `yarn ci:check` locally and it should show the same error as https://github.com/Nozbe/WatermelonDB/pull/932/checks?check_run_id=1765466338

jspizziri:
@radex, thanks for the feedback. I can clean up that failure. However, the test that I wrote doesn't actually validate that this particular problem is fixed, and I'm not entirely sure how to replicate it in tests with mocks since it involved the bridge. Do you have thoughts on this?

jspizziri:
@radex , I could use your help to figure out why creating the `mockDatabaseSQLite` seems to becausing all the SQL tests to fail. Any thoughts?

jspizziri:
@radex just wanted to ping you on this. Thanks!

radex:
@jspizziri I apologize for stalling for so long. I hope I didn't completely discourage you from contributing ever again...

I didn't answer because I couldn't figure out the right approach to tackle the test and had to take a look at it myself, and only now did I find the time to do so.

If you're curious, this is the approach I took in the end: https://github.com/Nozbe/WatermelonDB/commit/fa6ed89eb5f2e22809ad0f908ef1f073175fbea2#diff-fb928e10d2b89fe6541a2ef1ecc3282375fb2797e0e6e29ab2184ab75536fc82R90

jspizziri:
@radex not discouraged at all! I know you've got a lot of hopeful-contributors competing for your time.

Thanks for guiding me through getting this in. Will it be in a patch version anytime soon?

radex:
probably not, but I regularly release prereleases, and despite not being officially supported, they tend to be stable (we run them in production)

PR #933, opened 2021-01-26
Android File Path

avinashlng1080:
Provides the user with the ability to pass in a file path instead of the database name only and creates the *.db file under a 'databases' sub-folder under the provided path.

### DIFF
--- a/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
@@ -15,8 +15,21 @@ class Database(private val name: String, private val context: Context) {
                 if (name == ":memory:" || name.contains("mode=memory")) {
                     context.cacheDir.delete()
                     File(context.cacheDir, name).path
+                } else if (name.contains("/") || name.contains("file")) {
+                    // Extracts the database name from the path
+                    val dbName = name.substringAfterLast("/")
+
+                    // Extracts the real path where the *.db file will be created
+                    val truePath = name.substringAfterLast("file://").substringBeforeLast("/")
+
+                    // Creates the directory
+                    val fileObj = File(truePath, "databases")
+                    fileObj.mkdir()
+
+
+                    File("${truePath}/databases", dbName).path
                 } else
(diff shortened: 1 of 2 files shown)


native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -15,8 +15,21 @@ class Database(private val name: String, private val context: Context) {
                 if (name == ":memory:" || name.contains("mode=memory")) {
                     context.cacheDir.delete()
                     File(context.cacheDir, name).path
+                } else if (name.contains("/") || name.contains("file")) {
```
radex, line 18: i think we want it to start with `/` or start with `file:`, no?

radex:
@rozPierog can you review in a spare moment?

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
avinashlng1080: The file path is` file:///data/user/0/com.mattermost.rnbeta/files/xxx.db`.

Let me change the `contains` to `startsWith`

radex:
@rozPierog ⬆️

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -15,8 +15,21 @@ class Database(private val name: String, private val context: Context) {
                 if (name == ":memory:" || name.contains("mode=memory")) {
                     context.cacheDir.delete()
                     File(context.cacheDir, name).path
+                } else if (name.startsWith("/") || name.startsWith("file")) {
```
rozPierog, line 18: I feel like in this case we should force user to add `*.db` extension to the file. 
Can you add a check for that and maybe some throw with verbose error so that it's clear what failed?

rozPierog (changes requested):

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
avinashlng1080: done

avinashlng1080:
@rozPierog  kindly review this PR :)

Issue #958, opened 2021-03-08
Android Autolinking Docs & Breaking Change


jspizziri:
It appears that the recent implementation of Android autolinking has:

1. Left outdated install instructions. (WIP PR with fix https://github.com/Nozbe/WatermelonDB/pull/959)
2. Introduced a BC Breaking change (at least it did in our project). (PR with fix https://github.com/Nozbe/WatermelonDB/pull/960)

The work done on https://github.com/Nozbe/WatermelonDB/pull/805 and https://github.com/Nozbe/WatermelonDB/pull/884 introduced autolinking for Android. However, documentation [which was added to the `CHANGELOG` should've been added to the install documentation](https://github.com/Nozbe/WatermelonDB/pull/805/files#diff-bd3b90ea5e059997a16d0b05acfe64b7eeb85096cb700fc14fe7a76f0f8f1002R13). Additionally, the detail contained in the comment seems to have been removed by [a later commit which updated the `CHANGELOG`](https://github.com/Nozbe/WatermelonDB/pull/791/commits/813bb6b770f4499e5cc287c24daf716412088200#diff-bd3b90ea5e059997a16d0b05acfe64b7eeb85096cb700fc14fe7a76f0f8f1002L16). Lastly, the change appears to be a breaking one for anyone who had manually installed WatermelonDB prior to autolinking upon upgrade (at least in our project).

The Breaking Change error we're seeing in our project is the following:

```
Fatal Exception: java.lang.IllegalStateException
Native module DatabaseBridge tried to override DatabaseBridge. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
```

I have submitted two separate PR's one to address the install documentation and the other to address the Breaking Change in the `CHANGELOG`.

radex:
Closing in favor of only tracking your pending PR, thank you! ❤️

PR #959, opened 2021-03-08
WIP: docs(android): update the android installation instructions in light …

jspizziri:
…of autolinking

https://github.com/Nozbe/WatermelonDB/issues/958

### DIFF
--- a/docs-master/Installation.md
+++ b/docs-master/Installation.md
@@ -81,24 +81,13 @@ npm install @nozbe/with-observables
 
    See instructions above ⬆️
 
-2. In `android/settings.gradle`, add:
-
-   ```gradle
-   include ':watermelondb'
-   project(':watermelondb').projectDir =
-       new File(rootProject.projectDir, '../node_modules/@nozbe/watermelondb/native/android')
-   ```
-3. In `android/app/build.gradle`, add:
+2. In `android/app/build.gradle`, add:
    ```gradle
    apply plugin: "com.android.application"
    apply plugin: 'kotlin-android'  // ⬅️ This!
    // ...
-   dependencies {
-       // ...
-       implementation project(':watermelondb')  // ⬅️ This!
-   }
    ```
-4. In `android/build.gradle`, add Kotlin support to the project:
+3. In `android/build.gradle`, add Kotlin support to the project:
    ```gradle
    buildscript {
        ext.kotlin_version = '1.3.21'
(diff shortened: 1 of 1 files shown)


jspizziri:
So... In theory I think this should be right? However, this seems to break Android release builds with the following error:

```
> Task :nozbe_watermelondb:compileReleaseKotlin FAILED
e: /app/node_modules/@nozbe/watermelondb/native/android/src/main/java/com/nozbe/watermelondb/DatabaseDriver.kt: (47, 36): Unresolved reference: BuildConfig
```

I'm not sure why that'd be the case as the [Android Docs](https://developer.android.com/studio/build/gradle-tips#share-custom-fields-and-resource-values-with-your-app-code) state that the `BuildConfig` class is generated at build-time.

radex:
@rozPierog take a look pls, does this look right?

rozPierog:
I'm not familiar with how React Native handles auto linking to be honest. It seems strange that it would interfere with such fundamental thing as BuildConfig, but I welcome someone more knowledgeable on this to chime in

zxcpoiu (approved):
This is correct.

To use this lib, user should have setup babel config and add kotlin support in build.gradle, then the rest will be handled by auto linking. ( RN >= 60+ and WatermelonDB >= 21+ )

zxcpoiu:
FWIW: [how does autolinking works](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) if anyone interested in.

zxcpoiu:
@jspizziri have you tried to clear all gradle build cache? ie: `./gradlew clean`

jspizziri:
@zxcpoiu thanks for the feedback!

1. Yes, I did run a clean, I have it built in as part of my release command:
    ```json
    "android:clean": "cd android && ./gradlew clean",
    "android:release": "npm run android:clean && cd android && ./gradlew bundleRelease",
    ```
2.  I have `babel.config.js` as follows in my project:
    ```js
    module.exports = {
       presets: ['module:metro-react-native-babel-preset'],
       plugins: [
         ['@babel/plugin-proposal-decorators', { 'legacy': true }]
       ],
    };
    ```

zxcpoiu:
@jspizziri  I don't have this issue, it seems strange. :(

Anyway, I've research a bit and come up with a clearer solution and doc. Had Created a new PR #977 on top of cherry-picked your commit.

radex:
Closing in favor of #977 -- thank you @jspizziri  & @zxcpoiu

PR #971, opened 2021-03-24
GH Actions & updated native testing infra

radex:


native/iosTest/Podfile.lock
```diff
@@ -1,16 +1,380 @@
 PODS:
```
radex, line 1: CocoaPods bullshit  starts here - ignore until next comment

native/iosTest/WatermelonTester/AppDelegate.swift
```diff
@@ -8,21 +8,21 @@ class AppDelegate: NSObject, UIApplicationDelegate {
```
radex, line 8: … and we're back

radex:
I updated the whole native testing infra through more than a year of missed updates. So new RN, new pods, most new NPM packages… a lot of things exploded in our testing infra due to these updates, so I had to fix that… and then GH Actions. `c:check` works perfectly. iOS works fine, but slow (needs proper CocoaPods caching & ccache), Android works locally (after many battles), but requires more work to get the emulator running etc.

PR #973, opened 2021-03-25
Various changes, general maintenance…

radex:


radex:
@rozPierog Delegating this to you again, because the most interesting stuff is in GitHub actions setup. I managed to get the times for iOS down to 7 minutes and Android down to 11 by using @actions/cache for Yarn and CocoaPods & ccache to speed up Objc/C++ compilation. Not quite as good as self-hosted, but should be good enough for the foreseeable future.

.github/workflows/ci.yml
```diff
@@ -25,34 +25,41 @@ jobs:
...
+    - name: 'pod install'
+      if: steps.pods-cache.outputs.cache-hit != 'true'
+      run: cd native/iosTest && bundle exec pod install
     - run: yarn test:ios
     - run: yarn swiftlint
   android:
     # https://github.com/ReactiveCircus/android-emulator-runner/issues/46
     runs-on: macos-10.15
-    name: Check Android
+    name: Android tests
```
rozPierog, line 62: You can get Android to build even faster with caching - take Compose Sample app as a blueprint https://github.com/android/compose-samples/actions/runs/27015993/workflow

docs-master/Advanced/ProTips.md
```diff
@@ -2,14 +2,14 @@
 
 ## Database viewer
 
-See this discussion: https://github.com/Nozbe/WatermelonDB/issues/710
+[See discussion](https://github.com/Nozbe/WatermelonDB/issues/710)
 
-**Android** - you can use the new database inspector which comes with the Android Studio Beta. https://medium.com/androiddevelopers/database-inspector-9e91aa265316  .
+**Android** - you can use the [new database inspector](https://medium.com/androiddevelopers/database-inspector-9e91aa265316) which comes with the Android Studio Beta.
```
rozPierog, line 8: Database inspector is no longer only in beta branch. (also on newer versions Android Studio 4.2+ it's called `App Inspector` as it allows much more than only checking database)

rozPierog (changes requested):

.github/workflows/ci.yml
radex: noted for the future

PR #977, opened 2021-03-30
android: refine auto linking and doc

zxcpoiu:
This is a rework of #959 and #881.

## Motivation

Although on 0.21, the auto linking works, but user still have to manually edit their build.gradle, and the installation doc is a bit outdated.

This PR support real auto linking, means that user does not have to edit their native `build.gradle` files anymore.

Refers to [how react-native-webview handles kotlin version](https://github.com/react-native-webview/react-native-webview/blob/master/android/build.gradle)

## Explain:

In `build.gradle`, the `buildscript` block evaluated first even before the global var definition, and can specify dependency classpath there before building ( the `kotlin gradle plugin` ), so user don't have to specify it in rootProject.

This also is the reason why introducing `gradle.properties` file, since we can't access global vars in `buildscript` block yet.

If user has use `kotlinVersion` ext prop in root project, they can choose to use their own kotlin version avoiding to download different versions unnecessarily.

## Note:

* I changed the kotlin version name from `kotlin_version` to more conventional `kotlinVersion`, given that we are using `kotlin_version` before, and now we support true auto linking without user to specify, I don't think it is a breaking changes.

* I don't change the default android sdk version (currently 28) and kotlin version ( currently 1.3.21), maybe we can change the default to sdk 29 and kotlin 1.3.72, I leave it for maintainer's choice. Let me know if you want me to bump it as well.

### DIFF
--- a/docs-master/Installation.md
+++ b/docs-master/Installation.md
@@ -119,11 +119,11 @@ On RN60+, auto linking should work.
 
 <details>
   <summary>Custom Kotlin Version</summary>
-  Just set ext properties `kotlinVersion` in `android/build.gradle`
+  Just set ext properties `kotlinVersion` in `android/build.gradle`, and WatermelonDB will use the specified kotlin version.
   
   ```gradle
   buildscript {
-      ext.kotlin_version = '1.3.21'
+      ext.kotlinVersion = '1.3.21'
   }
   ```
 </details>
--- a/docs-master/Installation.md
+++ b/docs-master/Installation.md
@@ -75,45 +75,26 @@ npm install @nozbe/watermelondb
 
 ### Android (React Native)
 
-1. **Set up Babel config in your project**
+**Set up Babel config in your project**
 
-   See instructions above ⬆️
+See instructions above ⬆️  
 
-2. In `android/app/build.gradle`, add:
-   ```gradle
-   apply plugin: "com.android.application"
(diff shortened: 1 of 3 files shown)


docs-master/Installation.md
```diff
@@ -75,56 +75,67 @@ npm install @nozbe/watermelondb
...
+  ```
+</details>
+
+<details>
+  <summary>Custom Kotlin Version</summary>
+  Just set ext properties `kotlinVersion` in `android/build.gradle`
+  
+  ```gradle
+  buildscript {
+      ext.kotlin_version = '1.3.21'
```
radex, line 126: not kotlinVersion?

radex (changes requested):

radex:
@zxcpoiu Just a question

docs-master/Installation.md
zxcpoiu: Yeah, my bad, thanks for the eagle eyes :)
fixed

zxcpoiu:
Fixed, let me know if you want me to do a rebase

radex:
nah, I don't do rebases. LGTM

radex:
will merge after CI turns green

zxcpoiu:
Duplicate of #

radex:
@zxcpoiu hm?

zxcpoiu:
Sorry, my little monster touched my keyboard sneakily :sob:

PR #983, opened 2021-04-02
Sync: makes `pushChanges` optional

sidferreira:
The idea is to skip all the `localChanges` processing for databases that are "pull-only"

### DIFF
--- a/docs-master/Advanced/Sync.md
+++ b/docs-master/Advanced/Sync.md
@@ -9,7 +9,7 @@ Note that Watermelon is only a local database — you need to **bring your own b
 
 ## Using `synchronize()` in your app
 
-To synchronize, you need to pass two functions, `pullChanges` and `pushChanges` that talk to your backend and are compatible with Watermelon Sync Protocol. The frontend code will look something like this:
+To synchronize, you need to pass `pullChanges` and `pushChanges` _(optional)_ that talk to your backend and are compatible with Watermelon Sync Protocol. The frontend code will look something like this:
 
 ```js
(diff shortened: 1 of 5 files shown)


sidferreira:
cc @radex

src/sync/impl/synchronize.js
```diff
@@ -84,6 +84,10 @@ export default async function synchronize({
   }, 'sync-synchronize-apply')
 
   // push phase
+  if (!pushChanges) {
```
radex, line 87: I think the if statement should be scoped differently so that we get to the final logs (finishedAt, phase=done), which are still worth noting.

docs-master/Advanced/Sync.md
```diff
@@ -9,7 +9,7 @@ Note that Watermelon is only a local database — you need to **bring your own b
 
 ## Using `synchronize()` in your app
 
-To synchronize, you need to pass two functions, `pullChanges` and `pushChanges` that talk to your backend and are compatible with Watermelon Sync Protocol. The frontend code will look something like this:
+To synchronize, you need to pass `pullChanges` and `pushChanges` _(optional)_ that talk to your backend and are compatible with Watermelon Sync Protocol. The frontend code will look something like this:
```
radex, line 12: change log note pls

radex (changes requested):

src/sync/impl/synchronize.js
sidferreira: I missed that part indeed!
Addressed.

radex:
LGTM, thanks

PR #985, opened 2021-04-06
Fix, enable & document JSI use on Android

radex:


henrymoulton:
@radex is caching Cocoapods proving difficult?

I read this guide before https://dev.to/retyui/react-native-how-speed-up-ios-build-4x-using-cache-pods-597c might help

radex:
@henrymoulton Thanks for the link — TBH I'm not entirely comfortable with the complicated approach linked… I think for now we can just live with long iOS builds :)

rozPierog (approved):
Android java/kotlin/gradle code seems fine 🎉 I'm not comfortable enough with c++ code to evaluate it. 
Ping me in Nozbe if you are sure about merging this

PR #991, opened 2021-04-13
update typings for withDatabase()

zhirzh:
#### add static hoisting

```ts
class Foo extends Component {
  static foo() {}
}

const Baz = withDatabase(Foo)

Baz.foo()  // TS Error: Property 'foo' does not exist
```

#### optional requirement on `database` prop

```ts
type Props = {
  foo: number
}

class Foo extends Component<Props> {}

export default withDatabase(Foo)  // TS Error: Property 'foo' is missing in type '{ database?: Database | undefined; }' but required in type 'Readonly<{ foo: number; }>'.
```

#### allow `Database` subclass as prop

```ts
class MyDatabase extends Database {
  foo() {} // extra class member required to trigger TS error
}

type Props = {
  database: MyDatabase
}

class Foo extends Component<Props> {}

export default withDatabase(Foo)  // TS Error: Property 'foo' is missing in type '{ database?: Database | undefined; }' but required in type 'Readonly<{ foo: number; }>'.  // TS Error: Property 'foo' is missing in type 'Database' but required in type 'MyDatabase'

```

### DIFF
--- a/src/DatabaseProvider/withDatabase.js
+++ b/src/DatabaseProvider/withDatabase.js
@@ -1,8 +1,8 @@
 // @flow
 import React from 'react'
+import hoistNonReactStatics from 'hoist-non-react-statics'
 import type Database from '../Database'
 import { DatabaseConsumer } from './DatabaseContext'
-import hoistNonReactStatics from 'hoist-non-react-statics'
 
 type WithDatabaseProps<T: {}> = {
   ...T,
--- a/src/DatabaseProvider/withDatabase.js
+++ b/src/DatabaseProvider/withDatabase.js
@@ -2,6 +2,7 @@
 import React from 'react'
 import type Database from '../Database'
 import { DatabaseConsumer } from './DatabaseContext'
+import hoistNonReactStatics from 'hoist-non-react-statics'
 
 type WithDatabaseProps<T: {}> = {
   ...T,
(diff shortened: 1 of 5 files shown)


radex:
@zhirzh can you add note to changelog-unreleased?

src/DatabaseProvider/withDatabase.js
```diff
@@ -2,6 +2,7 @@
 import React from 'react'
 import type Database from '../Database'
 import { DatabaseConsumer } from './DatabaseContext'
+import hoistNonReactStatics from 'hoist-non-react-statics'
```
radex, line 5: why is this necessary?

zhirzh:
if I use `withDatabse()` on a component with some static members, I cannot access the statics on the returned component.

```ts
class Foo extends Component {
  static open() {}
}

const Baz = withDatabase(Foo)

Baz.open()  // Uncaught TypeError: Baz.open is not a function
```
This exact issue is fixed by copying my defined statics onto the component returned from `withDatabse()` and that is exactly what [`hoist-non-react-statics`][1] is used for.

[1]: https://github.com/mridgway/hoist-non-react-statics

radex:
@zhirzh OK. There's a CI failure though. You can reproduce locally by `yarn ci:check`

PR #1005, opened 2021-04-22
If there's nothing to delete, do not delete

sidferreira:
I'm using views in my DB and it tries to delete nothing on those views and causing errors

### DIFF
--- a/src/sync/impl/markAsSynced.js
+++ b/src/sync/impl/markAsSynced.js
@@ -40,7 +40,7 @@ const destroyDeletedRecords = (db: Database, { changes }: SyncLocalChanges): Pro
     mapObj(
       ({ deleted }, tableName) => deleted.length
         ? db.adapter.destroyDeletedRecords(tableName, deleted)
-        : Promise.resolve([]),
+        : Promise.resolve(),
       changes,
     ),
   )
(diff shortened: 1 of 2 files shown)


radex:
do you have these error messages on hand? This shouldn't cause any errors

sidferreira:
@radex I did the change on github website... will fix them in the next minutes

sidferreira:
@radex 

```
[logging] cannot modify my_table because it is a view in "delete from `my_table` where id in ()"
DB Error: 1 "cannot modify my_table because it is a view"
DB Query: delete from `my_table` where id in ()
```

this query runs after `sync` for EVERY table, even if 0 records

sidferreira:
I still wanna save some time and propose a per-table control about being part of sync or not.
In our case, we have basically 2 different sources of sync :)

radex:
OK, the patch seems reasonable. Do you wanna add a simple test to check that adapter is not being called in this case? (hint: spy + check number of calls to an adapter method) - I can see this easily getting reverted by accident at some point

sidferreira:
> OK, the patch seems reasonable. Do you wanna add a simple test to check that adapter is not being called in this case? (hint: spy + check number of calls to an adapter method) - I can see this easily getting reverted by accident at some point

will do it later today

sidferreira:
@radex done

radex:
Looks solid, thanks!

Issue #1027, opened 2021-05-10
Do not synchronize record or table


osvaldokalvaitir:
I am facing two synchronization problems.

1 - I have a table that serves to store application data. I can't use Watermelon's LocalStorage. How can I prevent this table from being synchronized.

2 - I have an order page that cannot be sent to the server until the user has entered all the information. The user can close the application and continue with the current order, it happens that until he finishes I cannot send it to the server. The solution I would have at the moment would be to make a 'new_order' table to create and then move to the 'order', but I would have to create a relationship for two tables, which I don't think is right, the best would be if I had a status so that I could ignore this record.

I found this problem in the issue, but I don't know what the solution was: https://github.com/Nozbe/WatermelonDB/issues/206

sidferreira:
I believe if we had a way to overwrite 
```
const createdQuery = Q.where(columnName('_status'), 'created')
const updatedQuery = Q.where(columnName('_status'), 'updated')
```

for `fetchLocal` is a start...

osvaldokalvaitir:
The test did not work with `order._raw._status = 'draft'`.

After a while, the status automatically changes to 'created'.

Until I have a good solution, I will duplicate the order table and I will prevent this table from being synchronized.

osvaldokalvaitir:
@sidferreira thanks for dedicating your time to my problem =)

KrisLau:
> 2 - I have an order page that cannot be sent to the server until the user has entered all the information. The user can close the application and continue with the current order, it happens that until he finishes I cannot send it to the server. The solution I would have at the moment would be to make a 'new_order' table to create and then move to the 'order', but I would have to create a relationship for two tables, which I don't think is right, the best would be if I had a status so that I could ignore this record.

I have something similar to this right now in my application. The user has go through a couple pages before the creation process is complete. I basically just either send the partially completed data through the navigation params or I save the data to `AsyncStorage` and fetch it later when needed and clear it after the record is created

osvaldokalvaitir:
> > 2 - I have an order page that cannot be sent to the server until the user has entered all the information. The user can close the application and continue with the current order, it happens that until he finishes I cannot send it to the server. The solution I would have at the moment would be to make a 'new_order' table to create and then move to the 'order', but I would have to create a relationship for two tables, which I don't think is right, the best would be if I had a status so that I could ignore this record.
> 
[...]

PR #1029, opened 2021-05-11
Adds `pushChanges` queries for sync customization

sidferreira:
Based on #1027

The suggestion is to add extra parameters to `pushChanges` so people can customize the sync process.

> Note: It required more changes than I expected and I'm not sure I'm happy with it.

### DIFF
--- a/src/Collection/index.js
+++ b/src/Collection/index.js
@@ -83,6 +83,10 @@ export default class Collection<Record: Model> {
     return new Query(this, clauses)
   }
 
+  queryWithDeleted(...clauses: Clause[]): Query<Record> {
+    return new Query(this, clauses, false)
+  }
+
   // Creates a new record in this collection
   // Pass a function to set attributes of the record.
   //
--- a/src/sync/impl/fetchLocal.js
+++ b/src/sync/impl/fetchLocal.js
@@ -11,6 +11,7 @@ import {
 import allPromisesObj from '../../utils/fp/allPromisesObj'
 import type { Database, Collection, Model } from '../..'
 import * as Q from '../../QueryDescription'
+import type { Clause } from '../../QueryDescription'
 import { columnName } from '../../Schema'
 
 import type { SyncTableChangeSet, SyncLocalChanges } from '../index'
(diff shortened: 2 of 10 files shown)


src/Collection/index.js
```diff
@@ -83,6 +83,10 @@ export default class Collection<Record: Model> {
     return new Query(this, clauses)
   }
 
+  queryWithDeleted(...clauses: Clause[]): Query<Record> {
```
radex, line 86: what legitimate use cases are there for fetching deleted records?

radex:
@sidferreira Can you elaborate how this new API is supposed to be used - how users would be to use it exactly? If there really is a large legitimate use case for many different ways of customizing sync queries, fine, but because of performance and correctness concerns I mentioned above, I'm hesitant to add a generic extension point like this if something simpler can solve the issue.

What I understood from @osvaldokalvaitir in #1027 is that there exists a need to make certain tables non-synchronizable (which can be done in a simpler way) and certain records created-but-not-yet-ready-to-synchronize, which also sounds like a legit improvement to Watermelon Sync that could be achieved by formalizing a new `draft` sync status without a generic customization point.

WDYT?

osvaldokalvaitir:
> @sidferreira Can you elaborate how this new API is supposed to be used - how users would be to use it exactly? If there really is a large legitimate use case for many different ways of customizing sync queries, fine, but because of performance and correctness concerns I mentioned above, I'm hesitant to add a generic extension point like this if something simpler can solve the issue.
> 
> What I understood from @osvaldokalvaitir in #1027 is that there exists a need to make certain tables non-synchronizable (which can be done in a simpler way) and certain records created-but-not-yet-ready-to-synchronize, which also sounds like a legit improvement to Watermelon Sync that could be achieved by formalizing a new `draft` sync status without a generic customization point.
> 
> WDYT?

exact this is what i need =)

src/sync/impl/fetchLocal.js
```diff
@@ -11,45 +11,77 @@ import {
...
+async function fetchLocalChangesForCollection<T: Model>({
+  collection,
+  createdQueries = [],
+  updatedQueries = [],
+  deletedQueries = [],
+}: fetchLocalChangesForCollectionArgs<T>): Promise<[SyncTableChangeSet, T[]]> {
   const [createdRecords, updatedRecords, deletedRecords] = await Promise.all([
-    collection.query(createdQuery).fetch(),
-    collection.query(updatedQuery).fetch(),
-    collection.database.adapter.getDeletedRecords(collection.table),
```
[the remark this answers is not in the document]
sidferreira: Yes, I totally agree with you on this point and that's why I mentioned `It required more changes than I expected and I'm not sure I'm happy with it.`. 

At the same time, I think we should have options to query deleted especially to avoid `collection.database.adapter.getDeletedRecords(collection.table)`, that basically breaks the `model => collection => database` structure (part of the decoupling I mentioned).

Maybe changing to `collection.query(updatedQuery).fetchIds()` to be a very specific method.

I'm not happy with this solution, especially on the part that skips the cache, but I think there are too many hard-coded rules (`createdQuery`, `updatedQuery`) that make sense to be customizable IMHO.

In short, I'm not happy with the purposed solution, but I think it's worth thinking about the concept of customization.

Even the `draft` state might be troublesome, as some users may want the `drafts` to be synced. An email app would like `drafts` to be synced. And in this case, they would need another `status` field.

I believe that #1027 starts a discussion about `pullable, pushable, local-only, read-only` scenarios.

sidferreira:
@radex btw, I don't have the needs that @osvaldokalvaitir described, I just purposed something based on what I felt when I put on his shoes.

`draft` does solve this issue, but maybe we can think a bit more broadly and have more options.

src/Collection/index.js
sidferreira: @radex undelete?!

radex:
@sidferreira hey, any reason why you closed this and another PR of yours?

sidferreira:
@radex it was marked as won't fix... I just assumed it wasn't relevant...

radex:
no, no, it's was just a rogue bot - I appreciate your PRs, and while I can't get to properly reviewing them now, they are useful

PR #1034, opened 2021-05-19
Q.unsafeSqlQuery()

radex:
Introduces `Q.unsafeSqlQuery` and deprecates `Collection.unsafeFetchRecordsWithSQL`. This has a number of benefits:

- Make query configuration API consistent and remove a weirdo API on Collection
- SQL queries can now be made safer thanks to SQL placeholders (`?`)
- You can now make unsafe SQL _count_ queries
- You can now observe SQL queries (using query reloading strategy)
- In the near future, I will introduce a new `adapter.queryIds` API (it will be needed for efficient delete implementation), and again, using a standard way to define queries means that one will be able to make raw SQL queries for Ids.


native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -41,9 +44,35 @@ class Database(private val name: String, private val context: Context) {
 
     fun delete(query: SQL, args: QueryArgs) = db.execSQL(query, args)
 
-    fun rawQuery(query: SQL, args: RawQueryArgs = emptyArray()): Cursor = db.rawQuery(query, args)
+    fun rawQuery(sql: SQL, args: QueryArgs = emptyArray()): Cursor {
```
radex, line 47: @rozPierog Do you mind taking a look at this function? Maybe something stupid I did jumps at you...

src/QueryDescription/index.js
```diff
@@ -256,15 +270,19 @@ export function where(left: ColumnName, valueOrComparison: Value | Comparison):
 }
 
 export function unsafeSqlExpr(sql: string): SqlExpr {
-  invariant(typeof sql === 'string', 'Value passed to Q.unsafeSqlExpr is not a string')
+  if (process.env.NODE_ENV !== 'production') {
```
michalpopek, line 273: Why only not in production? Shouldn't this trigger a runtime exception?

src/QueryDescription/index.js
```diff
@@ -256,15 +270,19 @@ export function where(left: ColumnName, valueOrComparison: Value | Comparison):
...
+  }
   return { type: 'sql', expr: sql }
 }
 
 export function unsafeLokiExpr(expr: any): LokiExpr {
-  invariant(
-    expr && typeof expr === 'object' && !Array.isArray(expr),
-    'Value passed to Q.unsafeLokiExpr is not an object',
-  )
+  if (process.env.NODE_ENV !== 'production') {
```
michalpopek, line 280: Ditto

michalpopek (approved):

src/QueryDescription/index.js
radex: I figured that is unlikely to get triggered outside of development - it would because you misused the API by accident. If someone passes unsanitized SQLs to Q (not a string literal), then a) it will just throw on query, b) you likely have bigger issues anyway. Runtime checks affect code size and runtime performance, so generally they're all wrapped in production check in WatermelonDB so they can be stripped entirely for production...

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -41,9 +44,35 @@ class Database(private val name: String, private val context: Context) {
...
+                        query.bindString(i + 1, arg)
+                    } else if (arg is Boolean) {
+                        query.bindLong(i + 1, if (arg) 1 else 0)
+                    } else if (arg is Double) {
+                        query.bindDouble(i + 1, arg)
+                    } else if (arg == null) {
+                        query.bindNull(i + 1)
+                    } else {
+                        throw (Throwable("Bad query arg type"))
+                    }
```
rozPierog, line 68:
```suggestion
                    when (arg) {
                        is String -> query.bindString(i + 1, arg)
                        is Boolean -> query.bindLong(i + 1, if (arg) 1 else 0)
                        is Double -> query.bindDouble(i + 1, arg)
                        null -> query.bindNull(i + 1)
                        else -> throw (Throwable("Bad query arg type"))
                    }
```

rozPierog (changes requested):
Seems fine, but I propose cleaning it up

PR #1035, opened 2021-05-20
Query cleanup + introduce Query.fetchIds()

radex:
- `Queries with multiple top-level `Q.on('table', ...)` now produce a warning. Use `Q.on('table', [condition1, condition2, ...])` syntax instead.` - previously there was a piece of magic that compressed multiple Q.ons into one, but that adds complexity and often produced suboptimal results. Instead, we're deprecating the magic, warning the users to fix up their queries, and letting advanced users optimize queries by arranging them in an optimal order
- Clean up Node.js implementation of SQLiteAdapter
- Add adapter.queryIds(), and matching query.fetchIds(). Fetching IDs only is significantly faster than fetching whole records, so it's an optimization opportunity. Also, this will be required for a a new performant implementation of record tree deletion. Aaalso, I'm going to delete adapter.findDeletedRecords(), since queryIds() is a generalized version of the same thing, and this isn't so performance critical as to require a custom micro optimized version.


src/adapters/sqlite/encodeQuery/index.js
```diff
@@ -209,6 +209,7 @@ const encodeLimitOffset = (limit: ?number, offset: ?number) => {
 const encodeQuery = (query: SerializedQuery, countMode: boolean = false): [SQL, SQLiteArg[]] => {
   const { table, description, associations } = query
 
+  // TODO: Test if encoding a `select id.x` query speeds up querIds() calls
```
michalpopek, line 212:
```suggestion
  // TODO: Test if encoding a `select id.x` query speeds up queryIds() calls
```

michalpopek (approved):
LGTM

PR #1036, opened 2021-05-20
SQLiteAdapter native implementations cleanup - part 1

radex:
The following is inspired by a few ongoing things:

- I'll need to be able to more efficiently delete many records at once for new delete API
- We now have FOUR different SQLite native implementations (Swift, Kotlin, C++/JSI and JS/Node), which is annoying to maintain and even more annoying when changes or new features are needed. I don't expect Swift and Kotlin implementations to be able to go away in 2021 (We probably need ubiquitous JSI in RN - probably only after Fabric ships; and we need an asynchronicity story - probably requiring Concurrent React to ship), and more might come in the future - e.g. Electron. So we want the tiniest possible API between JS and native land (but as performant as possible)
- OSS people want to modify Watermelon's behavior in ways I did not design for, e.g. calling into question why adapter's getDeletedRecords/destroyDeletedRecords are even a thing and not something that can be done using generic tools - see e.g #1029
- #1024 intends to introduce more stuff to the native implementations because of performance, so I'd like to remove other stuff in return ;) 

Therefore:

- getDeletedRecords is gone from adapters' native sides in favor of reusing queryIds
- setLocal/removeLocal are gone in favor of reusing batch(execute)
- remove some very old native iOS tests - those are completely superseded by JS bridge tests and redundant
- remove a big annoying iOS dependency (Nimble - tool for native tests) - not needed anymore, whatever was left was rewritten to use built-in XCTest framework


michalpopek (approved):
LGTM

PR #1037, opened 2021-05-21
SQLiteAdapter native implementations cleanup - part 2

radex:
continuing from https://github.com/Nozbe/WatermelonDB/pull/1036 :

- The internal batch() interface between JS portion of sqlite adapter and the native side has been redesigned.
- It's a little wonky, I admit, but it's very simple, really. Instead of encoding a list of fixed create/update/delete operations, we encode a SQL, and cache behavior (whether native adapter should mark record ID as cached/not cached), and an array of argument arrays
- First of all, moves the responsibility for generating all SQLs to JS, which is nice for code reuse (you can see that the only reason why the LOC delta is positive is because we have more tests, all adapters have shrunk)
- Second of all, the new operation encoding is tighter and more efficient - good for performance
- Third of all, instead of generating a single SQL per a record to create/update/delete, we group operations by the type (table and kind), generate a SQL once, and then have a list of argument lists - this is very good for performance
- I set up a benchmarking environment, and adapter.batch() is now between 16% and 48% faster (depending on ios/android/jsi and create/update/delete) 🎉
- This also paves the way for future optimizations, such as reusing sqlite prepared statement, and deleting&recreating sqlite indexes for very large batches of inserts/updates (if N is large enough, it's faster to remove the index and then recreate it from scratch than to update it incrementally)


src/adapters/sqlite/encodeBatch/index.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // $FlowFixMe
+    currentOperation[2].push(rawOrId)
+  })
+  if (currentOperation) {
+    grouppedOperations.push(currentOperation)
+  }
+  return grouppedOperations
+}
```
michalpopek, line 87: IMO reducing `operations` with a `[groupedOps, currentOp]` accumulator could be a bit more cleaner

src/adapters/sqlite/index.js
```diff
@@ -286,17 +264,18 @@ export default class SQLiteAdapter implements DatabaseAdapter {
     this._batch(
       [
         [
-          'execute',
+          0,
```
michalpopek, line 267: An descriptive enum (eg. `CacheBehavior.Ignore`) might help newcomers to grasp what's going on here without the need to delve into the implementation

michalpopek (approved):

src/adapters/sqlite/encodeBatch/index.js
radex: probably, but this is performance-critical

src/adapters/sqlite/index.js
radex: I did something similar to your suggestion in #1040… I don't like the exported enum pattern in JavaScript because it unnecessarily adds to the code size :/ But I defined locally named constants to help the reader understand what's going on.

src/adapters/sqlite/encodeBatch/index.js
radex: fixed in #1040

src/adapters/sqlite/encodeBatch/index.js
```diff
@@ -0,0 +1,129 @@
...
+  args[1] = raw._changed
+  for (let i = 0; i < len; i++) {
+    args[i + 2] = raw[(columns[i].name: any)]
+  }
+  args[len + 2] = raw.id
+
+  return args
+}
+
+type GrouppedBatchOperation =
```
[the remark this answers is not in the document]
radex: fixed in #1040

PR #1040, opened 2021-05-24
adapter.unsafeExecute & lokijs cleanup

radex:
- Introduced `adapter.unsafeExecute` to complete the "ability to drop down to metal" story, as suggested in #1039
- Fixed longstanding bug where adapter.setLocal would not be validated to be a string, causing differences between platforms
- Major cleanup and rearrangement of adapter tests, and lokijs. The internal naming of lokijs files has been changed to be roughly the same as in sqlite - since dispatcher/bridge/driver split is very similar to what lokijs already had


src/__tests__/databaseTests.js
```diff
@@ -1,858 +1,877 @@
+// @flow
```
radex, line 1: Enable "hide whitespace" to view this PR's diff:

image

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -53,18 +53,13 @@ class Database(private val name: String, private val context: Context) {
...
+                for ((i, arg) in args.withIndex()) {
+                    when (arg) {
+                        is String -> query.bindString(i + 1, arg)
+                        is Boolean -> query.bindLong(i + 1, if (arg) 1 else 0)
+                        is Double -> query.bindDouble(i + 1, arg)
+                        null -> query.bindNull(i + 1)
+                        else -> throw (Throwable("Bad query arg type"))
                     }
                 }
                 return SQLiteCursor(driver, editTable, query)
```
rozPierog, line 65: Sorry for getting in here again, but I've found more time to look at it again and I want to propose something looking even more like proper Kotlin code 

```Kotlin
        val rawArgs = Array(args.size) { "" }
        return db.rawQueryWithFactory(
            { _, driver: SQLiteCursorDriver?, editTable: String?, query: SQLiteQuery ->
                args.withIndex().forEach { (i, arg) ->
                    when (arg) {
                        is String -> query.bindString(i + 1, arg)
                        is Boolean -> query.bindLong(i + 1, if (arg) 1 else 0)
                        is Double -> query.bindDouble(i + 1, arg)
                        null -> query.bindNull(i + 1)
                        else -> throw IllegalArgumentException("Bad query arg type: ${arg::class.java.canonicalName}")
                    }
                }
                SQLiteCursor(driver, editTable, query)
            }, sql, rawArgs, null, null
        )
```
Most of that is just cosmetics, but I do think that giving name of bad type on error (and actually marking it as IllegalArgument) is going to add great deal to error readability

rozPierog (changes requested):

src/__tests__/databaseTests.js
```diff
@@ -1,858 +1,877 @@
...
+    matching: Array<{ id: string, ... }>,
+    nonMatching: Array<{ id: string, ... }>,
+    skipQuery?: boolean,
+    skipCount?: boolean,
+    skipLoki?: boolean,
+    skipSqlite?: boolean,
+    skipMatcher?: boolean,
+  }>,
+): void {
+  matchTests.push(options)
```
michalpopek, line 24: Why is `matchTest` and `joinTest` needed? Wouldn't providing an exact type for `matchTests` and `joinTests` instead of `any` do the job?

michalpopek (approved):

src/__tests__/databaseTests.js
radex: it's not _really_ needed, it's just that joinTests are skipped for `encodeMatcher` tests, since they're all incompatible. I can refactor it but doesn't really matter IMO ¯\_(ツ)_/¯  I don't want to spend time typing out tests, I only enabled flow in this one particular file to catch most obvious errors - missing fields for the tests and such

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
radex: applied in #1047 , thanks!

PR #1049, opened 2021-05-26
Android: avoids `.db.db`

sidferreira:
Breaking change?

### DIFF
--- a/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
@@ -15,12 +15,18 @@ class Database(private val name: String, private val context: Context) {
                 // TODO: This SUCKS. Seems like Android doesn't like sqlite `?mode=memory&cache=shared` mode. To avoid random breakages, save the file to /tmp, but this is slow.
                 // NOTE: This is because Android system SQLite is not compiled with SQLITE_USE_URI=1
                 // issue `PRAGMA cache=shared` query after connection when needed
+
                 if (name == ":memory:" || name.contains("mode=memory")) {
                     context.cacheDir.delete()
                     File(context.cacheDir, name).path
-                } else
+                } else {
+                    var sanitizedName = name
+                    if (!sanitizedName.endsWith(".db")) {
+                        sanitizedName = "$name.db"
+                    }
                     // On some systems there is some kind of lock on `/databases` folder ¯\_(ツ)_/¯
-                    context.getDatabasePath("$name.db").path.replace("/databases", ""),
+                    context.getDatabasePath(sanitizedName).path.replace("/databases", "")
+                },
                 null)
     }
 


native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -15,12 +15,18 @@ class Database(private val name: String, private val context: Context) {
                 // TODO: This SUCKS. Seems like Android doesn't like sqlite `?mode=memory&cache=shared` mode. To avoid random breakages, save the file to /tmp, but this is slow.
                 // NOTE: This is because Android system SQLite is not compiled with SQLITE_USE_URI=1
                 // issue `PRAGMA cache=shared` query after connection when needed
+
                 if (name == ":memory:" || name.contains("mode=memory")) {
                     context.cacheDir.delete()
                     File(context.cacheDir, name).path
-                } else
+                } else {
+                    var sanitizedName = name
```
radex, line 23: No, we can't accept this - this is a breaking change. I agree it would be better for the database to be properly suffixed, but it's not okay to just break people's apps. Why is database file name important to you?

sidferreira: @radex I think it's all about consistent behavior. 

Either both `iOS` and `Android` add the extension or neither do.

The docs mention we can set a `dbName` but there are no details about how the name will be changed. If one wants to, somehow, access the file, one must be able to properly be aware of the file name and path.

Happens that on iOS, if you specify a full path ( https://github.com/Nozbe/WatermelonDB/blob/33ce2962cd3804f3a435c7b805dc598a4fa71c07/native/ios/WatermelonDB/DatabaseDriver.swift#L225 ) it will not change the name at all. On the other hand, Android will always change the name, except if it is an in-memory database.

My suggestion is to add a similar rule to android (something I didn't accomplish yet in this PR)

radex: Truthfully, `dbName` mostly exists for backwards compatibility - it's only useful if you want to set the full path (or, i guess if you want to have multiple databases). Backwards compatibility is more important than consistency, unfortunately. If we were to break compatibility, we would have to rename the prop from dbName to something else and crash if the old one was used -- to ensure that users fix after the breaking change. 

Yes, setting the full path should work the same on Android.

Worth updating Setup.md so that users don't feel encouraged to set the name of the database unless they want to set a full path for native access

sidferreira: > Truthfully, `dbName` mostly exists for backwards compatibility - it's only useful if you want to set the full path (or, i guess if you want to have multiple databases). Backwards compatibility is more important than consistency, unfortunately. If we were to break compatibility, we would have to rename the prop from dbName to something else and crash if the old one was used -- to ensure that users fix after the breaking change.
> 
> Yes, setting the full path should work the same on Android.
> 
> Worth updating Setup.md so that users don't feel encouraged to set the name of the database unless they want to set a full path for native access

In this case, what about we add `databaseName` instead of `dbName`.
- `dbName` will translate into `databaseName`, but always adding `.db` if using android.
- `databaseName` will check if the extension is missing (maybe move this part to JS?)
- Ensure both handles the paths and names the same way (file db, memory db, etc)

if we're "forcing" single database usage, this conflicts with the discussion we had about "singleton databases".

I believe having the full path is also important to be able to ship a prepopulated database.

@radex

PR #1059, opened 2021-06-04
Turbo - part 5

radex:
Adds an optimization where large batch calls on sqlite use the age-old index recreation trick to speed up the operation.


src/adapters/sqlite/encodeSchema/test.js
```diff
@@ -52,18 +51,84 @@ describe('encodeSchema', () => {
...
+        'create index "tasks__status" on "tasks" ("_status");',
+    )
+  })
+})
 
-    expect(encodeSchema(testSchema)).toBe(expectedSchema)
+describe('encodeIndices', () => {
+  it(`encodes creation of indices`, () => {
+    expect(encodeCreateIndices(testSchema)).toBe(
+      '' +
```
michalpopek, line 76: Why does this need to be prepended to each query?

michalpopek (approved):

src/adapters/sqlite/encodeSchema/test.js
radex: oh it doesn't, it's a no-op to make Prettier format the query in an easier to read way, nothing to see here :)

PR #1066, opened 2021-06-10
Turbo - part 7

radex:
Taking Part 6 the extra mile.

The idea is that if we're processing the entire sync JSON in C++, why the heck are we pushing it from native networking code, to JS code, and then back to C++? On iOS, it's not _that_ big of a deal, but on Android, the cost of bridging between JVM and JS is significant, and I saw a nice performance bump when providing JSON from native to native, skipping JS altogether.

Also, a significant decrease in peak memory usage - also a very big deal on Android. Note that previously a 50MB JSON might have been copied multiple times before being used and freed -- and ballooned because JS is dumb UTF-16, which doubles the in-memory size.

So this PR introduces a native API that developers can use to `provide` a sync JSON with some unique id (a sequentially incrementing or a random number is ok), and then from JS, call unsafeLoadFromSync providing that ID.

For testing (and for less native-savvy OSS users), there's a JS api that you can call with a string to `provideSyncJson` - that then calls the native API. So there's a bunch of glue code, but it's all fairly straightforward.

One quirk of this implementation is that the management code for "provided sync JSONs" is implemented per-platform. The reason why is that native code will produce an NSData* on iOS or a byte[] on Android with the json - so to make this value usable until it's used, we either have to copy it (bad - memory usage, performance), or hold on to the platform native reference for as long as we need the data. I chose the latter. This way we can use the underlying raw data copy-free and let the platform do the final memory management.

Additionally, I added the recreate indices performance optimization here, and some extra tests and fixes.

### DIFF
--- a/native/android-jsi/src/main/cpp/DatabasePlatformAndroid.cpp
+++ b/native/android-jsi/src/main/cpp/DatabasePlatformAndroid.cpp
@@ -167,7 +167,7 @@ void provideJson(int id, jbyteArray array) {
     }
     assert(env);
 
-    if (providedSyncJsons.find(id) == providedSyncJsons.end()) {
+    if (providedSyncJsons.find(id) != providedSyncJsons.end()) {
         jclass exceptionClass = env->FindClass("java/lang/Exception");
         env->ThrowNew(exceptionClass, "sync json is already provided");
         return;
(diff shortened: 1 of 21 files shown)


michalpopek (approved):

marvanov:
Tell me what I did wrong?
`import { synchronize } from '@nozbe/watermelondb/sync'
export default async function watermelonSync(database) {
    await synchronize({
        database, unsafeTurbo: true,
        pullChanges: async () => {
            const response = await fetch(`https://marvanov.ru/api/wordskills/ru000.json`)
            if (!response.ok) return false
            const json = await response.text()
            return { syncJson: json } //! Error: provideSyncJson unavailable
        },
    })
    return true
}`
An error is thrown:
    at provideSyncJson (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.wordskills&modulesOnly=false&runModule=true:190845:27)

matheusgoncalves56:
i have the same error

radex:
did you pass `jsi: true` to SQLiteAdapter? @marvanov @matheusgoncalves56 ?

LovesWorking:
> SQLiteAdapter

TY!

PR #1069, opened 2021-06-14
Allow different id generator per schema

almassapargali:
Currently our tables use BigInt as id, and we use `setGenerator` to generate it ourselves. And now we'd like to use uuid as id on some tables, but `setGenerator` doesn't allow generating different ids per schema. I think solution to this would be passing schema name to randomid's generator. If you think that's the right approach, I'll add more code to this PR finalizing the feature.


radex:
I think it makes more sense to pass the whole TableSchema object than just the table name. This way, someone with even more elaborate needs than you could, for example, differentiate between multiple instances of a WatermelonDB Database, without moving a very simple helper function to be a property of Database (something that would surely be cleaner, but too complex for my taste in this particular case)

PR #1077, opened 2021-06-23
Subscription nesting fix

radex:
```
    // If we observe a list of records, and then for each of them we observe the record
     // We must emit list changes first - otherwise, a change to a record that removes it from the list
     // will re-render its component unnecessarily before the whole list is re-rendered to remove it
```

To achieve that, I needed to change the order of subscriber notifications - now, we first notify database subscribers, then collection subscribers, and model subscribers last. There's a good chance this will cause some regression somewhere else, but I haven't found anything so far.


kokusGr:
LGTM 👍

PR #1117, opened 2021-07-24
Fix tiny nit in Query documentation

olliewd:
I feel it's clearer with an await there as otherwise returns it a promise, not users.


radex:
@olliewd can you fix in docs-master folder? .html is generated from that

radex:
Thanks

PR #1118, opened 2021-07-26
Jsi threading fix

radex:


rozPierog (approved):
Tested this in NT and couldn't reproduce the issue 🙇

PR #1125, opened 2021-08-05
Fix typo in docs: `withComponents` -> `withObservables`

williamdclt:
`withComponents` doesn't exist, this is clearly about `withObservables` :)


radex:
oops, thanks!

PR #1129, opened 2021-08-10
Add link to Laravel Watermelon package

nathanheffley:
[Laravel Watermelon](https://github.com/nathanheffley/laravel-watermelon) is a Composer (PHP) package that provides an easy way to implement an endpoint to your [Laravel](https://github.com/laravel/laravel) project.

Currently, it only provides somewhat rudimentary sync functionality, but I am actively implementing the rest of the major features. Stuff left to do:

1) More robust authorization. There is a way to restrict users from viewing/updating/deleting records based on whatever authorization method you want, but if a user is authorized to see a record they can update any fields or delete the record. Per-field authorization and authorizing create/update/delete still needs implemented.

2) Schema Versioning is not implemented yet.

3) Migrations are not implemented yet.

So needless to say this package is still V1. I am hopeful its inclusion could still be valuable as a starting point for others to implement their own endpoint in other languages, and perhaps Watermelon DB users interested in Laravel could contribute to some of these missing features, but I totally understand if only fully-featured implementations are desired for this section.

### DIFF
--- a/docs-master/Advanced/Sync.md
+++ b/docs-master/Advanced/Sync.md
@@ -449,6 +449,7 @@ Note that those are not maintained by WatermelonDB, and we make no endorsements 
 
 - [How to Build WatermelonDB Sync Backend in Elixir](https://fahri.id/posts/how-to-build-watermelondb-sync-backend-in-elixir/)
 - [Firemelon](https://github.com/AliAllaf/firemelon)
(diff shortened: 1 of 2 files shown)


radex:
@nathanheffley Please update the Sync.md file in docs-master… the .html file is generated from that

nathanheffley:
Sorry about that, I've moved the change over to that file 👍

PR #1156, opened 2021-09-02
Database corruption - fix attempt 2137

radex:
- add the possibility of passing open flags to Database.kt
- add a secret debug hook for all database errors for NT use
- add a more understandable error when running methods on DB fails due to DB not having opened properly (previously this would show a mysterious `.apply` failure)
- add a debug hook for manually closing jsi database from JS

### DIFF
--- a/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
+++ b/native/android/src/main/java/com/nozbe/watermelondb/Database.kt
@@ -8,7 +8,11 @@ import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteQuery
 import java.io.File
 
-class Database(private val name: String, private val context: Context, private val openFlags: Int = SQLiteDatabase.CREATE_IF_NECESSARY or SQLiteDatabase.ENABLE_WRITE_AHEAD_LOGGING) {
+class Database(
+    private val name: String,
+    private val context: Context,
+    private val openFlags: Int = SQLiteDatabase.CREATE_IF_NECESSARY or SQLiteDatabase.ENABLE_WRITE_AHEAD_LOGGING
+) {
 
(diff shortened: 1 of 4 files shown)


radex:
@rozPierog ⬆️

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -8,20 +8,21 @@ import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteQuery
 import java.io.File
 
-class Database(private val name: String, private val context: Context) {
+class Database(private val name: String, private val context: Context, private val openFlags: Int = SQLiteDatabase.CREATE_IF_NECESSARY or SQLiteDatabase.ENABLE_WRITE_AHEAD_LOGGING) {
```
rozPierog, line 11: keep the code clean. Split that line (if you are using AS to edit this stuff just hit ⌥⌘L)

rozPierog (changes requested):

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
radex: meh, would be nice if prettier could reformat it as I went in VSCode, since AS is borderline unusable on M1, and ktlint didn't complain

radex: @rozPierog fixed

rozPierog: We can think about integrating some prettier plugin for KTlint. I'll look into this for Nozbe internal repo to see if we can easliy reuse same rules as ktlint uses'

> ktlint didn't complain

That's worrisome, and indicates that it might have broke

PR #1210, opened 2021-11-12
Update build.gradle

thestackdev:
Latest gradle v7.0.3 , recommended by android studio requires mininum kotlin v1.3.40


radex:
@rozPierog ?

PR #1220, opened 2021-12-01
Adds `skipIndexing` prop in TableSchema

sidferreira:
Our use of WatermelonDB includes some views, that can't be indexed.
v0.24 had issues with this use, so I'm adding this prop to keep our system running


sidferreira:
@radex maybe we can change `skipIndexing` to `isIndexed` (like the columns) and make it `true` by default (to avoid breaking change)

sidferreira:
@radex thinking about changing it to `isIndexed` with default to `true`...

Issue #1224, opened 2021-12-06
Random crash on startup - watermelondb::SqliteDb::destroy


mlecoq:
I have random crashes on startup of my app (after creating an account a new connection is made)

The dbName is user id 

I have the feeling that it deals with previously created db (with another user), I have `watermelondb::SqliteDb::destroy` in my logs

```
OS Version: Android 11 (RP1A.200720.012.T736BXXU1AUK4)
Report Version: 104

Exception Type: Unknown (SIGABRT)

Application Specific Information:
Abort

Thread 0 Crashed:
0   libc.so                         0x761e916ea4        abort
1   libc.so                         0x761e97fa64        <unknown> + 507319417444
2   libc.so                         0x761e97ef00        <unknown> + 507319414528
3   libc.so                         0x761e97edb4        pthread_mutex_lock
4   split_config.arm64_v8a.apk      0x750768fef4        sqlite3_next_stmt
5   split_config.arm64_v8a.apk      0x75077744a8        watermelondb::SqliteDb::destroy
6   split_config.arm64_v8a.apk      0x7507774624        watermelondb::SqliteDb::~SqliteDb
7   split_config.arm64_v8a.apk      0x75077828e8        <unknown> + 502636488936
8   split_config.arm64_v8a.apk      0x750777502c        watermelondb::Database::~Database
9   libhermes-executor-common-release.so0x7518219660        facebook::jsi::DecoratedHostObject::~DecoratedHostObjec

```

radex:
are you using the latest version of 🍉?

mlecoq:
yes, In fact it is related to https://github.com/Nozbe/WatermelonDB/issues/1052#issuecomment-851401260

When there is a switch of current account in my app (connection/logout then new connection)

```
(async () => {
        await new Promise(resolve => setTimeout(resolve, 0));
        setInitialized(true);
      })();
```

is not reexecuted (it's executed once at first connection)  

It was a bit hard to guess since this issue appears randomly (and not on every devices)

radex:
@mlecoq Are you doing any sort of React Native bridge reload or something like that? There is a crash on pthread-something, which reminds me of an issue I've already fixed that occurred on bridge reload - there would be an attempt to use a system thread that was finalized

mlecoq:
No there is no reload and the crash was generated on a built app bundle

It was systematic after user creation so that's why I suspect database connection (since my connection depends on user id)

mlecoq:
[...]

PR #1227, opened 2021-12-14
feat: expose dbName in adapter

mlecoq:
I have issues with managing several user (and several underlying db - see #1224), to avoid creating too many database I want to check that current one is related to connected user 

What I'm doing right now is:

```
if (
      user.status === 'connected' &&
      // @ts-ignore
      (!dbRef.current || dbRef.current.adapter.underlyingAdapter._dbName !== user.id)
    ) {
      dbRef.current = database(user.id);
    }
```

It's a bit low level and I would like to be able to get dbName directly from adapter

### DIFF
--- a/src/adapters/compat.js
+++ b/src/adapters/compat.js
@@ -25,6 +25,10 @@ export default class DatabaseAdapterCompat {
     return this.underlyingAdapter.schema
   }
 
+  get dbName(): ?string {
+    return this.underlyingAdapter._dbName
+  }
+
   get migrations(): ?SchemaMigrations {
     return this.underlyingAdapter.migrations
   }
--- a/src/adapters/compat.js
+++ b/src/adapters/compat.js
@@ -26,7 +26,7 @@ export default class DatabaseAdapterCompat {
   }
 
   get dbName(): ?string {
-    return this.underlyingAdapter._dbName
+    return this.underlyingAdapter.dbName
   }
 
   get migrations(): ?SchemaMigrations {
--- a/src/adapters/lokijs/test.js
+++ b/src/adapters/lokijs/test.js
@@ -9,8 +9,9 @@ import DatabaseAdapterCompat from '../compat'
 describe('LokiJSAdapter (Synchronous / Memory persistence)', () => {
   commonTests().forEach((testCase) => {
     const [name, test] = testCase
-    const dbName = `test${Math.random()}`
+
     it(name, async () => {
+      const dbName = `test${Math.random()}`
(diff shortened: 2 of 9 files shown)


src/adapters/compat.js
```diff
@@ -25,6 +25,10 @@ export default class DatabaseAdapterCompat {
     return this.underlyingAdapter.schema
   }
 
+  get dbName(): ?string {
+    return this.underlyingAdapter._dbName
```
radex, line 29: I'd prefer if the underlying adapters also had a `dbName`, not a `_dbName`. the compat layer is meant to turn callback-based functions into promise-returning functions.

src/adapters/lokijs/test.js
```diff
@@ -9,17 +9,18 @@ import DatabaseAdapterCompat from '../compat'
 describe('LokiJSAdapter (Synchronous / Memory persistence)', () => {
   commonTests().forEach((testCase) => {
     const [name, test] = testCase
-
+    const dbName = `test${Math.random()}`
```
radex, line 12: why? it's not wrong here per se, since this is memory-backed, but potentially error-prone since you don't want that in tests with actual persistent storage (or _shared_ memory storage as in sqlite)

radex:
@mlecoq Yeah that sounds good but please correct a few bits, check that CI passes, and add your change to CHANGELOG-Unreleased.md please :)

mlecoq:
@radex thanks for your feedback 

I will need more time to understand why integration tests are failing 

on Android I have 

```
03:46:11 V/InstrumentationResultParser: Expected: "file:testdb0.624348272712959-android"
1623
03:46:11 V/InstrumentationResultParser: Received: "file:testdb157?mode=memory&cache=shared"
```

I'm a bit confused since the first name is for async test and the second one for jsi

src/adapters/compat.js
mlecoq: I have renamed _dbName into dbName (see https://github.com/Nozbe/WatermelonDB/pull/1227/commits/cab09a2d87a11c872b729db28c59f8f90d050424)

src/adapters/lokijs/test.js
mlecoq: fixed: https://github.com/Nozbe/WatermelonDB/pull/1227/commits/61ecb6b94dfd2c28131652707bb4f86ee0ab70e6

mlecoq:
There is still a failure on android job but it is due to a network issue

radex:
lgtm, thanks

PR #1234, opened 2021-12-17
Added Blast Workout logo to README

alwex:
Adding Blast Workout logo to README.
Thank you for this fantastic library.


radex:
@alwex can you send a different crop so that the logo is not so much taller than the others?

PR #1327, opened 2022-05-17
feat: support sortBy on joined table

mlecoq:
Currently if we make a query with joined tables, we cannot sort on columns of joined tables. 

The goal of this PR is to be able to indicate on which table sort is applied


radex:
Looks nice, but please either add relevant support to `adapter/lokijs` too, or at least, add a check that will throw an Error when attempting to construct a loki query with this feature with a useful error message

radex:
please also add a test here: https://github.com/Nozbe/WatermelonDB/blob/master/src/__tests__/databaseTests.js#L860 that actually checks for correct behavior (not just expected query syntax). this is important for checking cross-adapter compatibility

mlecoq:
Thanks for your feedback

mlecoq:
> Looks nice, but please either add relevant support to `adapter/lokijs` too, or at least, add a check that will throw an Error when attempting to construct a loki query with this feature with a useful error message

I have chosen to throw an error, sorting on joined tables doesn't look to be supported currently on loki.js

KrisLau:
@radex Could you please take a look at #1252 ? It's a PR by an Expo team member

suman379:
Is this still active? I am waiting for this.

PR #1336, opened 2022-06-06
Make Android Database class single instance for safe use in the native calls

rozPierog:
- ⚠️ do not merge, needs to be tested on a live app first


native/android/src/main/java/com/nozbe/watermelondb/Database.kt
```diff
@@ -8,25 +8,40 @@ import android.database.sqlite.SQLiteDatabase
...
+        @JvmStatic
+        fun getInstance(
+            name: String,
+            context: Context,
+            openFlags: Int = SQLiteDatabase.CREATE_IF_NECESSARY or
+                SQLiteDatabase.ENABLE_WRITE_AHEAD_LOGGING
+        ): Database =
+            INSTANCE ?: synchronized(this) {
+                INSTANCE ?: buildDatabase(name, context, openFlags).also { INSTANCE = it }
             }
```
radex, line 25: Am I understanding this correctly - INSTANCE is a singleton global? i.e. whatever you pass to getInstance is only used if this is the first time that it's called, and otherwise you just get whatever was constructed before?

If I do, this is way too lazy and awfully bad to merge. Fine to test as a proof of concept (i'll make ya a prerelease in a moment), but this effectively would make it impossible to create multiple Database instances - would break tests and other people's apps.

Can't we just access React context from widget/etc, and from it, drill down to DatabaseBridge → driver → database? That would be the simplest and cleanest solution.

If this is not possible (because it's not necessary going to be initialized), then a cheapo but acceptable solution is that we pass to Database (all the way from JS) some string identifier (something like `unsafeNativeReuseId`) - if passed, this would assign the Database (or maybe DatabaseDriver?) with this id to a Map - or reuse if present; and some other global method (accessible from app Java code) could also be used to fetch or initialize a Database with this id.

radex (changes requested):

radex:
@rozPierog Check out `0.24.1-6`

native/android/src/main/java/com/nozbe/watermelondb/Database.kt
rozPierog: > Am I understanding this correctly - INSTANCE is a singleton global? i.e. whatever you pass to getInstance is only used if this is the first time that it's called, and otherwise you just get whatever was constructed before?

Yes.

> If I do, this is way too lazy and awfully bad to merge […] but this effectively would make it impossible to create multiple Database instances

I'm aware

> Can't we just access React context from widget/etc, and from it, drill down to DatabaseBridge → driver → database? That would be the simplest and cleanest solution.

That approach would invalidate any reason for this change. We want to have database access _without_ going trough anything related with react native. Sometimes, some parts of the app are just unaware that there _is_ something like ReactContext somewhere else.

> [last paragraph]

I had an idea to use dbName as this identifier. It needs to be unique anyway, it points to a db single file that should only be _opened once anyway_, so all in theory seem good. If you are against it then I guess we can add some another prop for that, but I would like to keep it simple.

radex: > I had an idea to use dbName as this identifier. It needs to be unique anyway, it points to a db single file that should only be opened once anyway, so all in theory seem good. If you are against it then I guess we can add some another prop for that, but I would like to keep it simple.

In theory you should be able to have multiple connections to a database. I'd feel more comfortable if this Database object caching only happened if requested. I think it's ok if we use dbName as an identifier. So… `_unsafeNativeReuse: true/false` ?

PR #1318, opened 2022-05-02
Typescript Bump

sidferreira:
There's a big gap between Flow and Typescript typings and this PR is trying to fix this difference.

Instead of using the current `modules` approach for TS, (almost) every JS file has its own `.d.ts` file.

From now on, `.d.ts` files needs to be updated only if a signature changes and it is easier to non TS engineers guess how to update it.

PR #1340, opened 2022-06-16
fix(ts): wrong migration arg type for pullChanges method

mlecoq:
In pullChanges arguments, currently migration is typed as a schema migration (with `toVersion` and `steps` attributes)

That is not consistent with what is actually returned : https://github.com/Nozbe/WatermelonDB/blob/4e7abb5ccbdccc018d657673ccc2c72ae06d0681/src/Schema/migrations/getSyncChanges/index.js#L11


radex:
@mlecoq Thank you for the PR, alas it got into conflict with @sidferreira's big Typescript changes. Can you resolve conflicts?

mlecoq:
@radex the PR #1318 from @sidferreira fixes this issue so mine is useless now.  @sidferreira , thanks a lot for your works on types

PR #1354, opened 2022-07-15
Update gradle.properties

joaodias-rms:
Updating the kotlin version to avoid errors


radex:
@rozPierog LGTY?

radex:
@rozPierog ping

PR #1430, opened 2022-11-04
Integrate MelonFire into list of sync implementations

fivecar:
I've implemented a sync for Firebase which handles issues with the current alternative (firemelon):

- MelonFire overcomes Firestore's 500-write transaction limit
- It doesn't get timestamps wrong (e.g. if the server timestamp isn't consistent)
- It follows WatermelonDB's sync protocol strictly (e.g. one retry, MUST/SHOULD actions all done, lazy deletes tracked, etc)
- It guarantees atomic reads (i.e. can survive multiple writers)


fivecar:
@radex : not sure if I might need your approval to be able to run the check-in workflow on this PR. Please LMK if I need to do anything else to unblock this! Thanks!

radex:
thanks, nice work!

PR #1440, opened 2022-11-23
Replace deprecated jcenter() repo with mavenCentral()

GEllickson-Hover:
JCenter has been deprecated a while and is read-only right now, but it has become an unreliable repository and can cause build failures intermittently from server issues when building without cache. There's been a couple times at least since deprecation that jcenter has had server issues for a full day (as I doubt they're spending much resources maintaining it anymore).

https://blog.gradle.org/jcenter-shutdown

In particular to this project, the `native/android/build.gradle` declaration is the most problematic as apps consuming the lib won't easily be able to override the this buildscript repo declaration. It can be safely replaced with mavenCentral() as the Kotlin plugin you're using it for exists there as well.

### DIFF
--- a/examples/native/android/build.gradle
+++ b/examples/native/android/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
     repositories {
         google()
-        jcenter()
+        mavenCentral()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


rozPierog:
I see no issues with that. Thanks!

PR #1447, opened 2022-12-07
I would like to add more docs about observable api.

lucaswitch:
The page can be found under advanced section: 
domain/Advanced/UsingObservables.html


lucaswitch:
@radex Can you take a look at this pull request?
 Then if this pull request is legit i can continue to develop this doc page.

PR #1449, opened 2022-12-08
docs: updated LocalStorage reference to be correct

fivecar:
The old docs referred to `localStorage`, but it would seem the latest LocalStorage functions are actually based off `db.adapter`. I updated the LocalStorage.md and then did `npm run docs`.

Note that I installed the latest version of mdbook (0.4.22), which likely led to all the other changes you see. Unless perhaps a previous run of mdbook was done manually by someone with custom settings.

Please LMK if instead I should be simply modifying the md file and not generating the docs... because I notice that some of the docs are updated in this diff with actual content (which I did not produce).


fivecar:
@radex : I don't believe these failures have to do with my change (e.g. by looking through the error logs), especially since I only updated an MD file and re-ran the doc-generation path. You can also see that my changes don't touch the errored files in question.

Could you let me know the best way to proceed?

radex:
This is not correct. The documentation is correct.

radex:
> What am I missing? It seems the shipped d.ts files on npm are different from what's currently shown in master.

Of course they are. `master` includes latest changes not yet shipped on NPM

fivecar:
@radex: It doesn't bother you that the documentation differs from what's actually usable via npm? It feels to me that the only point of documentation should be to make it possible to use a package...

radex:
The published documentation: https://nozbe.github.io/WatermelonDB/Advanced/LocalStorage.html matches the published code. The unpublished documentation on master matches unpublished code on master.

fivecar:
At the risk of sounding like even more of an idiot than I already do, I'm still definitely missing something. The published documentation that you linked:
image

The published 0.24 `d.ts` are, I believe, instead aligned with [what I mentioned above](https://github.com/Nozbe/WatermelonDB/pull/1449#issuecomment-1375835116).

This is what my VS Code looks like when I use the published package. Note no `localStorage` property:
image

This is the same db, now accessed via the method this PR proposes:
image

I must sincerely apologize for harping on this, because I half suspect you're going to patiently explain to me something incredibly dumb that I've done here. I realize it's painful to deal with people saying random things on open source projects, and so I thank you in advance for your patience in helping me understand my mistake.

radex:
@fivecar I'm sorry, I mis-read something. There IS an issue, but not where you originally thought it was.

- the 0.24 documentation does say you should use `database.localStorage` - I was wrong to imply otherwise
- but 0.24 DOES contain it! https://github.com/Nozbe/WatermelonDB/blob/v0.24.0/src/Database/index.js#L68

The problem isn't with the documentation - it's with 0.24 Typescript types - see https://github.com/Nozbe/WatermelonDB/blob/v0.24.0/src/Database/index.d.ts

However, there's nothing more to fix, except to publish a new version to remove the confusion - since 0.24, the kind contributors have improved Typescript typings, and they do contain `database.localStorage` https://github.com/Nozbe/WatermelonDB/blob/master/src/Database/index.d.ts#L42

fivecar:
Ah, ok, it all makes sense now. Thanks for explaining!

PR #1458, opened 2022-12-31
fix(doc): wrong import, field reference instead of immutableRelation

MarcosSarges:
In the documentation examples there is an import error. Field reference instead of immutableRelation

[Relation](https://nozbe.github.io/WatermelonDB/Relation.html)

Screenshot 2022-12-31 at 11 51 26


radex:
thanks

PR #1463, opened 2023-01-06
fix(sync): Remove records that were created and deleted in the same sync

### LINKED ISSUE #1462: Sync erroneously creates/updates records that should delete in same sync pass
REPRO:

1. Run an app that uses WatermelonDB and syncs to a server
2. Create record "aaa"
3. Sync
4. Delete record "aaa"
5. Sync
6. Delete app from phone
7. Reinstall app
8. Sync

RESULT:
Record "aaa" syncs down, and doesn't get deleted.

REASON:
When you sync with the fresh (empty) DB, the server sends down a `SyncDatabaseChangeSet` that includes `aaa` in its `created` and `deleted` arrays. Yet `applyRemoteChanges` doesn't delete `aaa`, and instead only deletes records that were already in the DB prior to the sync in Step 8 (which, for a fresh install, is zero records).

I've got a fix coming.

fivecar:
Essentially, the problem is that if `SyncDatabaseChangeSet` pulls down an `id` in created/updated that's also in deleted, the current code ends up creating that record anyway (and not deleting it). This makes some syncs (especially ones that start from an empty DB or haven't sync'd in a while) retain records that shouldn't actually be in the DB by the time sync finishes.

This fixes that by removing those records from the changeset before sync code erroneously creates/updates those records.

Note: I was not able to `yarn` this project, even from `master`, because the `npm:` style packages (e.g. npm:expect@24.1.0) don't resolve. Am I doing something wrong?

Tested and verified working in my own app. But I don't know how to test this, or even run automation, given the problem mentioned above. Would appreciate feedback on how to unblock that, and on this PR in general. (I'm a bit surprised this hasn't been reported/observed by others, so there's a fair chance I've simply misunderstood how things are supposed to work in sync).

fixes #1462

### DIFF
--- a/src/sync/impl/applyRemote.js
+++ b/src/sync/impl/applyRemote.js
@@ -238,7 +238,7 @@ const unsafeApplyAllRemoteChangesByBatches = (
 // deleted, we need to remove it from created so that the record doesn't end
 // up remaining in the DB.
 function reduceChanges(remoteChanges: SyncDatabaseChangeSet): SyncDatabaseChangeSet {
-  Object.keys(remoteChanges).forEach(tableName => {
+  Object.keys(remoteChanges).forEach((tableName: any) => {
     const changes = remoteChanges[tableName]
 
     changes.created = changes.created.filter(rec => !changes.deleted.includes(rec.id))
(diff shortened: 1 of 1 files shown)


fivecar:
(ooh - and a big thanks to whoever moved the codebase over to Typescript. Much easier.)

fivecar:
closing request, given discussion in #1462

PR #1476, opened 2023-01-21
Just some improvements to book.js

mikemikhaylov:


radex:
@mikemikhaylov Thanks so much for wanting to contribute. Unfortunately I can't accept this contribution :( This file is auto-generated by `mdbook`, so it doesn't make sense to touch it.

If you want to contribute around docs, see if you can upgrade `mdbook` to the latest version, since the current one got out of date.

Or better yet, see if you can migrate 🍉's docs to Docusaurus - it seems to be a standard tool in the React/RN community, and IMO looks much nicer than mdbook

PR #1498, opened 2023-01-27
kotlin 1.7.20

rozPierog:
Update dependencies, Kotlin version, fix deprecated calls


examples/native/android/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         minSdkVersion = 16
         compileSdkVersion = 29
         targetSdkVersion = 29
-        kotlin_version = '1.3.21'
+        kotlin_version = '1.7.21'
```
radex, line 9: This will require users to bump their kotlin too, right?

Issue #1481, opened 2023-01-24
Help migrate 🍉 docs from `mdbook` to Docusaurus


radex:
Hey,

I'd like to migrate WatermelonDB docs from `mdbook` to Docusaurus, a de facto standard in the React/React Native community for project websites. With that, I'd also like documentation to feature an actual API reference, and not just guides. I hope that will help 🍉 users a lot.

But with many other maintenance responsibilities on my mind, I'd like help from the community with this. Can you help?

KrisLau:
I can help but would probably just need some direction since I've never done this before!

ErickLuizA:
I want to help too, do you have anything in mind related to how it should be structured in the repo?

The current setup has the docs-master/ with the .md files and the generated docs/ folder.

Can it be changed to something like the structure found in the [react-native-navigation](https://github.com/wix/react-native-navigation/tree/master/website) repo, it features the API reference you mentioned or even [react-native-website](https://github.com/facebook/react-native-website)

Both of these examples also have versioning, is that something useful here?

radex:
@ErickLuizA Yes, I'd like to change the structure a bit, it seems to confuse a lot of people that docs-master/ is the current version of docs that matches code on master, while `docs/` matches the latest published version.

Regarding versioning, it's a nice to have, but not super important to me. I think we can start with a simple docusaurus setup, then add API reference, and then maybe think about how to ship versioned docs.

radex:
We have this! @ErickLuizA thank you for your contribution.

The new website will be shipped alongside 0.26.0 release, for now you can view it locally like so:

```
cd docs-website
yarn
cd ..
yarn docs:dev  # opens in browser
```

radex:
@KrisLau Your help would be definitely appreciated.

One thing you could start with if you're unsure about drafting new articles is just cleaning up the existing docs:

- Splitting large guide articles into smaller ones
- Throughout the docs, adding types to examples would be helpful. Things have changed since 2018, and I think we can now assume that most people use TypeScript/Flow, and whoever doesn't can just ignore them
- Changing navigation to be closer to your suggestion seems like a good idea

KrisLau:
@radex Will do!

radex:
[...]

PR #1507, opened 2023-01-31
Migrate docs docusaurus

ErickLuizA:
Ref #1481

This PR changes the documentation implementation from [mdbook](https://github.com/rust-lang/mdBook) to [docusaurus](https://github.com/facebook/docusaurus).

There are some points that I'm not too sure of and would like to discuss.

1. Assets -- like logo, favicon and etc...
2. Colors -- theme in the `src/css/custom.css`
3. Blog -- There is a "CHANGELOG" section currently, but I see a lot of websites using this blog feature as a changelog, maybe something to consider?
4. MDX -- Docusaurus uses mdx, which means that the markdown files have to be compatible with it, which implies stuff like needing to close tags (ex: `<br />` instead of `<br>`)
5. Deploy -- I have not tested the `deploy` script, I don't understand much about GitHub pages but don't think it would have much effect given the setup of a custom domain, the default behavior appears to be to create a `gh-pages` branch where it will spit the result of the `build` script.
6. Search -- Add search functionality to the docs(very, very useful!), the best option is apparently Algolia, here is the [Link ](https://docusaurus.io/docs/search#connecting-algolia)for setup, if it's not possible to use that, then, maybe, I can update with another search, there is several in the docs.

I have deleted the docs-master and docs folders, and also the mdbook scripts, don't know if that is right?


radex:
Awesome! Thanks @ErickLuizA - I'll try to take a look later today

radex:
Okay, this looks super sweet!

Here are some issues:

- `yarn docs:dev` gives me:

```
 ~/Documents/Nozbe/WatermelonDB   ErickLuizA-migrate-docs-docusaurus ⇕ yarn docs:dev
yarn run v1.22.4
$ ./scripts/docusaurus-dev
sed: 1: "./docs-website/docs/doc ...": invalid command code .
```

- it would be worth deduplicating the common part between `docusaurus` and `docusaurus-dev` scripts. I suggest making it into a common `script/update-docusaurus` and moving `yarn start/deploy` to package.json scripts
- let's bring back the `docs/` folder for now. It's where the official docs deploy from, and until 0.26 is released, it should stay there, as the new docusaurus docs will be WIP and contain info for newer versions of 🍉. Don't bring back docs-master, we'll keep the master docs with the docusaurus folder

and some other things, not at all blocking, but would be nice:

- do you know of a tool that would help create docusaurus-compatible API reference from the source code? It would be great to have an API reference, but I do not wish to maintain it separately from the codebase and keep it up to date. It would be easier if I could add extra markings in comments above functions and classes and have it synced with docs automatically.
- it would be great to have the "Version: xxxx" marker at the top of every page without having to opt into versioned docs. Actual versioning could be a pain to maintain, but people are sometimes confused when they look at "release" docs but use a prerelease/next version of 🍉

ErickLuizA:
The `sed` command works differently in Linux/macOS, so I switched to js and created the `replace-docs-path.mjs`, which is just to ensure the links are not broken anywhere.

About the version thing, I have a problem, I thought it would be good to get it from the `docs-website/package.json`, I created a `docs:version ` script in the main `package.json` and it works fine, however, that has the side-effect that the `docs` script cannot be run alone since the version needs to be updated before deploying.

I have added the `docs:version` to run in the `release.mjs`, don't know if that is where it's supposed to be? 

About the generate docs from source code, there are some solutions, I tested the [jsdoc-to-mdx](https://github.com/naver/jsdoc-to-mdx), it appears to work, although I kind of just put some text and see if it outputs something, if you wanna check it out, it's in this [branch](https://github.com/ErickLuizA/WatermelonDB/tree/migrate-docs-docusaurus-jsdoc), btw, for some reason, to work I need to put the comment in the `.js` and `d.ts` to work :thinking:

There is [this gist](https://gist.github.com/slorber/0bf8c8c8001505f0f99a062ac55bf442) from the maintainer of docusaurus, but I was not able to make it run, there is also [this](https://github.com/facebook/docusaurus/issues/3413), but it's for typescript.

radex:
Thanks @ErickLuizA! I'll merge this now, as it already looks better than what we have shipping.

>About the generate docs from source code, there are some solutions, I tested the [jsdoc-to-mdx](https://github.com/naver/jsdoc-to-mdx), it appears to work, although I kind of just put some text and see if it outputs something, if you wanna check it out, it's in this [branch](https://github.com/ErickLuizA/WatermelonDB/tree/migrate-docs-docusaurus-jsdoc), btw, for some reason, to work I need to put the comment in the .js and d.ts to work 🤔

This is very interesting and exciting, let's move that discussion to #1481 so we can merge this PR

PR #1477, opened 2023-01-23
🐛 Fix TypeScript  following #1318

paulrostorp:
This PR aims to revert breaking changes and broken types following #1318

PR #1509, opened 2023-01-31
🐛 TS: Fix wrong `DirtyRaw` type

paulrostorp:
see: https://github.com/Nozbe/WatermelonDB/pull/1477#issuecomment-1410077203

### DIFF
--- a/CHANGELOG-Unreleased.md
+++ b/CHANGELOG-Unreleased.md
@@ -30,7 +30,7 @@
 - [JSI] Improved reliability when reloading RCTBridge
 - [iOS] Fix "range of supported deployment targets" Xcode warning
 - [JSI] Improved reliability when reloading RCTBridge
-- [TypeScript] Fix type of `DirtyRaw` to `object` (from `Object`)
+- [TypeScript] Fix type of `DirtyRaw` to `{ [key: string]: any }` (from `Object`)
 
 ### Performance
 
--- a/CHANGELOG-Unreleased.md
+++ b/CHANGELOG-Unreleased.md
@@ -30,6 +30,7 @@
 - [JSI] Improved reliability when reloading RCTBridge
 - [iOS] Fix "range of supported deployment targets" Xcode warning
 - [JSI] Improved reliability when reloading RCTBridge
+- [TypeScript] Fix type of `DirtyRaw` to `object` (from `Object`)
 
 ### Performance
 
--- a/src/RawRecord/index.d.ts
+++ b/src/RawRecord/index.d.ts
@@ -5,7 +5,7 @@ import type { RecordId, SyncStatus } from '../Model'
 // Raw object representing a model record, coming from an untrusted source
 // (disk, sync, user data). Before it can be used to create a Model instance
 // it must be sanitized (with `sanitizedRaw`) into a RawRecord
-export type DirtyRaw = Object
+export type DirtyRaw = object
 
 // These fields are ALWAYS present in records of any collection.
 type _RawRecord = {
--- a/src/RawRecord/index.d.ts
+++ b/src/RawRecord/index.d.ts
@@ -5,7 +5,7 @@ import type { RecordId, SyncStatus } from '../Model'
 // Raw object representing a model record, coming from an untrusted source
 // (disk, sync, user data). Before it can be used to create a Model instance
 // it must be sanitized (with `sanitizedRaw`) into a RawRecord
-export type DirtyRaw = object
+export type DirtyRaw = { [key: string]: any }
 
 // These fields are ALWAYS present in records of any collection.
 type _RawRecord = {


radex:
@mlecoq Can you verify that this works?

mlecoq:
I have the same issue with object (with lowercase o)

`property '<PropertyName>' does not exist on type '{}'`

paulrostorp:
@mlecoq that is the intended effect 😄  see my comment [here](https://github.com/Nozbe/WatermelonDB/pull/1477#issuecomment-1410062220)

mlecoq:
@paulrostorp I saw your comment but I have 234 errors and checking every property will be very verbose 

It looks strange to have code which has to "guess" properties that are defined in schema 

I think that DirtyRaw should be derived from the schema 

In the meantime, requesting developers to compensate lost type informations by lots of checks seems annoying

radex:
> I think that DirtyRaw should be derived from the schema

It really shouldn't, as explained here: https://github.com/Nozbe/WatermelonDB/pull/1477#issuecomment-1410068421

Ideally, RawRecord could be schema-typed, i.e. `RawRecord<Task>`, etc.

DirtyRaw is basically untyped by design - its unknown type is exactly what it should convey

paulrostorp:
@mlecoq If you want to forego type safety, you can simply type cast it to the desired type, for example: Screenshot 2023-01-31 at 11 24 18
or
Screenshot 2023-01-31 at 11 25 35

The `object` type will the same signature as any object, so you don't need to cast to unknown first.


I don't believe there is an east way for the library to guess the type based on the schema here.
My personal opinion is that types should be strict and safe, otherwise why even have them, but this is a breaking change since #1318, @radex please advise

mlecoq:
@paulrostorp ok, i will cast all my sub methods of pushChanges. 

` don't believe there is an east way for the library to guess the type based on the schema here.`

you're right, it looks tricky and needs to look deeper in types. 


@radex, concerning `pushChanges` method, the changes param is:

```export type SyncTableChangeSet = $Exact<{
  created: DirtyRaw[];
  updated: DirtyRaw[];
  deleted: RecordId[];
}>```

You mean that it could be `RawRecord` instead of `DirtyRaw` ?

mlecoq:
As a matter of fact, if when we change the schema of database we could directly have side effects in pushChanges method (and notice regression eventually) it would be great

radex:
> You mean that it could be RawRecord instead of DirtyRaw ?

No, because this is user input, and we don't really know what the true types of these objects are.

If you fully trust that your backend gives you objects that have the correct type, then you can force casting these objects to the desired type. But it SHOULD raise an issue and it should be your explicit decision to silence that type error.

> you're right, it looks tricky and needs to look deeper in types.

it's doable, but it would require a major refactoring of how types work in 🍉, and would be a breaking change.

mlecoq:
@radex ok understood, so in fact the issue is that pullChange and pushChange methods use same types whereas for pushChanges types could be known : data is not from user input but from db, am I right ?

radex:
@mlecoq Yes, that  is correct - data passed to pushChanges is known to be type-safe

radex:
> @paulrostorp I've played around with it in TS playground, and I think `{ [id: string]: any }` is the right choice, at least for now

@paulrostorp As for long term, I think DirtyRaw maybe should be an opaque type, so that it's useful for 🍉 internals, but users would be forced to explicitly cast it into something untyped to remind them that DirtyRaw is dirty and not guaranteed to have correct types instead of the type-checker just accepting anything. But that would be a breaking change

paulrostorp:
@radex I'll revert to pre #1318 then.
But for the long term intent, then I would recommend the type to simply be `unknown` (https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown)

radex:
Yes, `unknown` is exactly what I meant )

mlecoq:
> @radex I'll revert to pre #1318 then.
> 
> But for the long term intent, then I would recommend the type to simply be `unknown` (https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown)

Unknown will generate another issues, I use graphql with graphql code generator so all my calls are typesafe: unknown cannot be directly cast to string, number ... so lot of casts in perspective

paulrostorp:
@mlecoq the whole point of this would to "generate issues". The point of type safety is to generate errors at compilation time instead of at runtime.

Typescript types have no effect at runtime. If I type a binary stream to an object type, it doesn't make it safe at runtime, your process will still crash !

paulrostorp:
I don't think it's feasible to infer types from the database schema (at least not without a significant rewrite of watermelonDB). I don't think you could transform the javascript object of the appSchema to typescript types without some custom scripting (like code generation tools); typescript alone cannot do this.

The closest thing to a type is your Model class, but nothing forces the database schema to follow that (for example, Date are stored as numbers), I think you are better off explicitly casting or typechecking.

radex:
Guys, please break up the fight :) Type-safety WILL be improved in the future (PRs are welcome), generating new errors, but here we just want to remove a clear regression.

@paulrostorp I'll merge it later today

radex:
Released in 0.25.4

radex:
contd. #1516

PR #1518, opened 2023-02-03
WatermelonDB Android Java implementation

rozPierog:
This PR changes android native code from Kotlin to Java, to simplify usage for new users, and to stop conflicts with kotlin versions of other libraries.


native/android/src/main/java/com/nozbe/watermelondb/WMDatabase.java
```diff
@@ -0,0 +1,191 @@
...
+    private WMDatabase(SQLiteDatabase db) {
+        this.db = db;
+    }
+
+    public static Map<String, WMDatabase> INSTANCES;
+
+    public static WMDatabase getInstance(String name, Context context, int openFlags) {
+        synchronized (WMDatabase.class) {
+            if (!INSTANCES.containsKey(name) ||
+                    !(INSTANCES.get(name) == null && INSTANCES.get(name).isOpen())) {
```
radex, line 25: this condition seems wrong

rozPierog: Duh, of course it's wrong
It should be `!= null`

radex: ok, fixed on another branch

PR #1519, opened 2023-02-03
Native cleanup

radex:


native/android-jsi/src/main/cpp/CMakeLists.txt
```diff
@@ -4,38 +4,54 @@ cmake_minimum_required(VERSION 3.4.1)
 # inspired by https://github.com/ericlewis/react-native-hostobject-demo/blob/6f16c01db80f928ccd294c8cc5d4668b0f8c15ec/android/app/CMakeLists.txt
 # execute_process (COMMAND ln "-s" "src" "../../../../../node_modules/react-native/third-party/double-conversion-1.1.6/double-conversion")
 
-# TODO: Fix duplication
```
radex, line 7: @rozPierog It's not _great_, but unlike my last attempt, I was able to quite easily clean up CMakeLists

PR #1535, opened 2023-02-15
Restructure sections

KrisLau:
As per described here: https://github.com/Nozbe/WatermelonDB/issues/1481#issuecomment-1404140038

I didn't add types as I don't use TypeScript and have no experience in it so I was worried about messing up haha


README.md
```diff
@@ -34,175 +34,3 @@
 | ✨ | **Reactive.** (Optional) [RxJS](https://github.com/ReactiveX/rxjs) API |
 | 🔗 | **Relational.** Built on rock-solid [SQLite](https://www.sqlite.org) foundation |
 | ⚠️ | **Static typing** with [Flow](https://flow.org) or [TypeScript](https://typescriptlang.org) |
-
```
radex, line 37: hmm, I'd like to keep README the way it was before - it's used on the GitHub page, not just on docs website

docs-website/docs/docs/CONTRIBUTING.md
```diff
@@ -3,22 +3,33 @@ title: Contributing
 hide_title: true
```
radex, line 3: This should be made in CONTRIBUTING file, as this one simply mirrors it

radex (changes requested):
Good start, however please fix a few issues before merge

docs-website/docs/docs/CONTRIBUTING.md
KrisLau: Do you mean I should make the changes only in the contributing file or in both? I'm assuming you mean both but I just want to make sure ahaha

README.md
radex: hmm, I'm not sure yet. I worry this would be an extra maintenance burden to have to update both the main README and the docs

docs-website/docs/docs/CONTRIBUTING.md
KrisLau: Do I need to run `yarn docs:dev` manually? or is that run as a part of the CI?  When I run it manually, it errors out

radex: Try `./scripts/update-docusaurus`. If doesn't work, what does it say?

KrisLau: ohh I just realized it is probably is because I ran it on Windows. The error is `'.\scripts\update-docusaurus' is not recognized as an internal or external command, operable program or batch file.`

README.md
KrisLau: @radex Sorry, just checked and realized that this is already in a separate file from the one used by the main README so I think it should be ok?

KrisLau: ok I'll revert this change! Maybe someone can figure out a script to copy based off the sections.  I tried asking around some forums and googling but to no avail ahha

KrisLau:
@ radex Updated with fixes

KrisLau:
@radex  Updated with fixes

radex:
thanks!

PR #1543, opened 2023-02-23
feat: add types to Model class about fetch method

ebnersilva:
This small PR fix the typescript problem when i try to get a relation in some model.

Example code: 

```
async function getUserRelation() {
  const userWorkCenterCollection = database.get<UsuarioCentroTrabalhoModel>(
    'user_work_centers',
  );

  const userWorkCenters = await userWorkCenterCollection
    .query(Q.where('work_center_id', data.value))
    .fetch();

  const user = await userWorkCenters[0].usuario.fetch(); // Here i was receiving a message error that do not have fetch method in Model class

  ...
}
```

### DIFF
--- a/src/Model/index.d.ts
+++ b/src/Model/index.d.ts
@@ -112,6 +112,8 @@ export default class Model {
 
   get table(): TableName<this>
 
+  public fetch: () => Promise<this>
+
   // FIX_TS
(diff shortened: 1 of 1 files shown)


ebnersilva:
Ok, i understand, but, why this code works?

This is the way i get a relation record!

PR #1548, opened 2023-02-27
Synchronization option now supports syncUpdateCondition to skip remote record

primus11:
Atm synchronization only allows to customize conflict resolution which doesn't provide full control when trying for example to skip remote record altogether. It is possible to use conflict resolution that would return local but this is happening too late and has side effect of modifying updated_at.


primus11:
I can move it yeah if that is preferred. My thinking at the end was that it maybe belongs to `requiresUpdate` and that if possible it is better that it is done before `sanitizedRow` and `areRecordsEqual` in it. Atm at least, it can also be before built-in `requiresUpdate`. Please just reconfirm and I will also change naming to `shouldUpdateRecord` at that point.

> BTW I'm not sure if this is correct. It's arguable whether this is expected behavior, but I think that's a bug. I specifically added doesn't touch created_at/updated_at when applying updates test

This puzzled me a bit as well. The thing is that `areRecordsEqual` is triggered before `conflictResolver` takes place which automatically means that this record will go into `prepareUpdate` at that point. My first thinking was to add control when `updated_at` is updated since I believe it could be used elsewhere but it would be a bit funky from `prepareUpdateFromRaw` perspective and this change just achieves wanted result a lot better.

radex:
> that if possible it is better that it is done before sanitizedRow and areRecordsEqual in it

Ok, rename it to `shouldUpdateRecord`, but keep the arguments as-is. We could export the original implementation (possibly renamed for consistency) so that one can use that.

src/sync/impl/__tests__/applyRemote.test.js
```diff
@@ -506,4 +506,86 @@ describe('applyRemoteChanges', () => {
...
+
+      await expectSyncedAndMatches(tasks, 'tUpdated', {
+        _status: 'updated',
+        _changed: 'name,position',
+        name: 'remote local', // concat of remote and local change
+        position: 100,
+        description: 'remote', // remote change
+        project_id: 'orig', // unchanged
+      })
+    })
```
primus11, line 589:
```
    it('forced local should keep record intact', async () => {
      const { database, comments } = makeDatabase()

      await makeLocalChanges(database)
      await testApplyRemoteChanges(database, {
        mock_comments: {
          updated: [{ id: 'cSynced', body: 'remote' }],
        },
      }, {
        conflictResolver: (
          tableName: TableName<any>,
          local: DirtyRaw,
          remote: DirtyRaw,
          resolved: DirtyRaw,
        ) => {
          return local;
        }
      })

      await expectSyncedAndMatches(comments, 'cSynced', {
        _status: 'synced',
        _changed: '',
      })
    })
```

will cause

```
 FAIL  src/sync/impl/__tests__/applyRemote.test.js (32.296 s)
   applyRemoteChanges  conflictResolver  forced local should keep record intact

    expect(received).toMatchObject(expected)

    - Expected  - 2
    + Received  + 2

      Object {
    -   "_changed": "",
    -   "_status": "synced",
    +   "_changed": "updated_at",
    +   "_status": "updated",
```

PR #1585, opened 2023-04-25
Fix out of date types of src/sync/impl/applyRemote

mfbx9da4:


radex:
@mfbx9da4 BTW how did you come across this issue? Do you use applyRemote directly?

mfbx9da4:
Yes, we don't use synchronize because it's not well suited to websockets and not well suited to sync-pushing part of a record so I've implemented something a bit more sophisticated but it still makes use of `_changes` and `applyRemoteChanges`.

PR #1654, opened 2023-08-31
feat: add renameColumn migration step

zhenkaGo:
I added the renameColumn migration step.


radex:
Thanks for this! Have you looked at https://github.com/Nozbe/WatermelonDB/pull/1211 ? IIRC, there's a similar problem here as with deleting columns - that it may not work for older versions of SQLite. So at the very least there should be a big fat warning about this, or a workaround that works for all versions

radex:
Thanks @zhenkaGo -- I'll close this PR as the work continues in https://github.com/Nozbe/WatermelonDB/pull/1799

PR #1661, opened 2023-09-07
chore: update invariant message

sebastian-schlecht:
Update invariant message to emit table name and id of the record to be updated to diagnose errors.


radex:
Thanks Seb!

PR #1671, opened 2023-09-24
Added destroyColumn migration step

bmatasar:
The initial PR was too old and syncing everything was a lot of pain, so I dropped it completely. The initial PR was #1211.


KrisLau:
@radex

Issue #527, opened 2019-10-18
[DB] Uh-oh. Database failed to load, we're in big trouble TypeError: Cannot read property 'version' of undefined


isllyend:
"react-native": "^0.59.3",
"@nozbe/watermelondb": "^0.14.1",

what's up? Please help me ):

chrise86:
I just faced this issue, what is the cause?

scarlac:
@chrise86 This happens when your schema doesn't exist, for example you've imported your model/schema.js file incorrectly. Check your `import schema from '...'` statement and ensure that it's the default import if using this particular syntax, or use `import { schema } from '...'` if you are using a named export.

PR #1692, opened 2023-10-21
Added missing schema error

scarlac:
This addresses the cryptic onboarding issue caused due to docs giving example of a default import, yet [showing code for a schema with a named export](https://watermelondb.dev/docs/Schema#defining-a-schema):
#527

(obv. fixing the docs would also be nice, but this error is nice to have regardless)


radex:
this diff did not pass CI but I added an equivalent check on master - thanks!

PR #1704, opened 2023-11-10
Update CRUD.md: Code sample

AlexOConnorHub:
Added a more complete code sample to "Create a new record" section. I spent a lot of time trying to figure out how to write in the database. After realizing there were tests, this is what I came up with based on the model tests.


AlexOConnorHub:
Closing as, upon reading comments in the class that models extend, this is NOT how you're supposed to do this.

PR #1706, opened 2023-11-14
fix: sqlite-node initialize error

### LINKED ISSUE #1705: In a Node.js environment, throws a resource lock error.
My basic code is as follows.

```javascript
import { mySchema } from './model/schema.js'

import SQLiteAdapter from "@nozbe/watermelondb/adapters/sqlite/index.js";

const adapter = new SQLiteAdapter({
    dbName: 'test',
    schema: mySchema,
    onSetUpError: error => {
        // Database failed to load -- offer the user to reload the app or log out
        console.log(error)
    }
})
```
At this point, running the code produces the following results.
image
How should it be handled?

KotoriMinami:
If this instance is not closed, it will result in the occupation of the database file. 
This leads to calling the `unsafeDestroyEverything` method during `setUpWithSchema`.
Although the connection of the new instance is closed, 
an error `resource busy or locked` still occurs when executing `fs.unlinkSync`.
This fixes #1705

### DIFF
--- a/src/adapters/sqlite/sqlite-node/DatabaseBridge.js
+++ b/src/adapters/sqlite/sqlite-node/DatabaseBridge.js
@@ -38,6 +38,9 @@ class DatabaseBridge {
       resolve({ code: 'ok' })
     } catch (error) {
       if (driver && error.type === 'SchemaNeededError') {
+        if (driver.database.instance && driver.database.instance.open) {
+          driver.database.instance.close()
+        }
         this.waiting(tag, driver)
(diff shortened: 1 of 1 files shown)


GitMurf:
@KotoriMinami is the reason you closed this PR is to replace it in favor of this? #1707

KotoriMinami:
> @KotoriMinami is the reason you closed this PR is to replace it in favor of this? #1707

Yes, since I'm not very familiar with using the PR feature, and I found that the current submission wasn't thorough enough, I created a new one.

PR #1707, opened 2023-11-15
fix:The database connection is not closed during the catch block

KotoriMinami:
If this instance is not closed, it will result in the occupation of the database file.

When subsequent calls to `setUpWithMigrations` or `setUpWithSchema` reach the execution of `unsafeDestroyEverything`, 
even though the connection of the new instance is closed, an error `resource busy or locked` still occurs when executing `fs.unlinkSync`."

This fixes https://github.com/Nozbe/WatermelonDB/issues/1705


radex:
Thanks for the fix @KotoriMinami, but tests no longer pass after this

PR #1749, opened 2024-02-08
Add missing bracket to code snippet

DiNitride:
withDatabase code snippet is missing a bracket for the compose function. PR adds this to the docs to avoid minor head scratching.

Some docs on what compose is and does would be nice too!


radex:
good catch, but the proposed diff isn't quite right -- I fixed it on master

PR #1758, opened 2024-03-11
Tauri adapter

ospfranco:
I was tasked to create an adapter for Tauri. Here is my attempt at it. Since the Tauri communication bridge between JS and Rust is async, I pretty much took the Node adapter and promisified every operation. Along with some other changes.

Code is not 100% ready to be merged as I wasn't sure what's the best way to expose this driver (as some of the calls are hardcoded for node). But at least I thought leaving it here doesn't hurt.


src/adapters/sqlite/makeDispatcher/index.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 /* eslint-disable global-require */
 
-import DatabaseBridge from '../sqlite-node/DatabaseBridge'
```
ospfranco, line 4: This path is hardcoded for node and I just replaced it for now.

PR #1808, opened 2024-06-26
Add json decorator options to TS type

martyfitzer-ids:
Modifies the json property decorator TS type to support an options object as a third function argument that the [JS implementation](https://github.com/Nozbe/WatermelonDB/blob/master/src/decorators/json/index.js#L41) supports.

### DIFF
--- a/src/decorators/json/index.d.ts
+++ b/src/decorators/json/index.d.ts
@@ -3,6 +3,11 @@ import Model from '../../Model'
 
 export type Sanitizer = (source: any, model?: Model) => any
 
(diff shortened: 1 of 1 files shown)


martyfitzer-ids:
Hi @radex, do you have any idea why the [windows tests check](https://github.com/Nozbe/WatermelonDB/pull/1808/checks#step:9:49) failed?

radex:
@martyfitzer-ids I don't know for sure, but don't worry about it -- it was broken before your changes. Seems like a Visual Studio update on GitHub's side broke it. I may have to bump React Native to a newer version for Windows CI to work again.

radex:
thanks for the TS improvement!

PR #1846, opened 2024-10-21
Add model name and raw data info to diagnostic errors

pranavbabu:
This PR adds model name and raw data of record to diagnostic errors to improve debugging processes.

Related to #612


radex:
Sorry @pranavbabu, this reveals too much information to the console to be the default -- but if you use the debug switch on Database (or add a new one), this seems like it could be very useful

radex:
see https://github.com/Nozbe/WatermelonDB/commit/3af430b94ae5dab44111939c773459dc83b7142b

PR #1854, opened 2024-11-16
Add jsi support for newer react native versions(old-arch)

### LINKED ISSUE #1833: jsimodule not working in 0.74.3 version
Hi i am using watermelondb after upgrade react native version 0.74.3 compile error showing.After some research in react native blog they removed jsimodule.

How to fix this issue? @radex

prathameshmm02:
Fixes #1833. 

Convert WMDB JSI Module to NativeModule which removes the need of overriding `getJSIModulePackage`.

This works for bare react native apps with versions 0.74-0.76 and Expo 51 as of now and does not work with Expo 52, bc of autolinking issues, I am looking into that.

### DIFF
--- a/docs-website/docs/docs/Installation.mdx
+++ b/docs-website/docs/docs/Installation.mdx
@@ -1,5 +1,5 @@
-import Tabs from '@theme/Tabs'
-import TabItem from '@theme/TabItem'
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 
 # Installation
 
@@ -167,7 +167,7 @@ additional steps manually.
 
     ```
     // ...
-    import com.nozbe.watermelondb.jsi.WatermelonDBJSIPackage // ⬅️ This!
+    import com.nozbe.watermelondb.jsi.WatermelonDBJSIPackage; // ⬅️ This!
     // ...
 
     @Override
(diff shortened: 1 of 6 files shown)


radex:
@prathameshmm02 Excellent work. I will need to test it in more detail before I can merge it into mainline, but if this approach works, it clearly seems like it would cause fewer installation troubles.

Is this something that other JSI libraries did as well?

Have you tested on relatively old RN versions? I believe the reason why I took the JSIModule approach originally instead of the one you took is that native modules were ran on the wrong thread and the whole thing would crash.

prathameshmm02:
> Is this something that other JSI libraries did as well recently?

Yeah, react-native-mmkv, reanimated & vision camera all had this approach, as they all use JSI on old arch.

> Have you tested on relatively old RN versions?

No, just 74 and 76 as of now, it should work (probably) though.

> I believe the reason why I took the JSIModule approach originally instead of the one you took is that native modules were ran on the wrong thread and the whole thing would crash.

`@ReactMethod(isBlockingSynchronousMethod = true)` this makes it work as it's synchronous.

Also, it would be better if installation with Expo 52 with new [autolinking](https://docs.expo.dev/modules/autolinking/) would be fixed, as there are a lot of Expo users(including me). The issue with autolinking is that android & ios folders of the project are in native folder, though, in my testing this has only caused issue on android, ios works fine.

radex:
@prathameshmm02 Thank you, this was helpful!

Issue #1282, opened 2022-03-13
_raw is always empty


amhed:
I'm running into an issue similar to these ones: 
- https://github.com/Nozbe/WatermelonDB/issues/104
- https://github.com/Nozbe/WatermelonDB/issues/682

When using `create` the write operation succeeds into writing, but the object is always written with empty values. Inspecting the return object I can see that `_raw` has all the fields defined by the schema, but they all have the default values instead.

schema
```ts
import { appSchema, tableSchema } from '@nozbe/watermelondb';

export const schema = appSchema({
  version: 5,
  tables: [
    tableSchema({
      name: 'wallet',
      columns: [
        { name: 'walletId', type: 'string', isIndexed: true },
        { name: 'currencyCodeStr', type: 'string', isIndexed: true },
      ],
    }),
  ],
});
```

Model
```ts
import { Model } from '@nozbe/watermelondb';
import { field } from '@nozbe/watermelondb/decorators';

export class WalletWMO extends Model {
  static table = 'wallet';

  @field('walletId')
  walletId!: string;

  @field('currencyCodeStr')
  currencyCodeStr!: string;
}
```

Code to insert
```ts
database.write(async () => {
  const table = database.get<WalletWMO>('wallet');
  const result = await table.create((record: WalletWMO) => {
    record.walletId = this.id;
    record.currencyCodeStr = this.currencyCode.rawValue;
  };);
});
```

And yet, the underlying `_raw values are empty/default

<img width="259" alt="image" src="https://user-images.githubusercontent.com/579291/158071367-9b558519-efbd-4a13-9bd6-8d77ad3bedd7.png">

Tsconfig has 
```
"experimentalDecorators": true,
"emitDecoratorMetadata": true,
```

BabelConfig has
```
const plugins = [
    'react-hot-loader/babel',
    ['@babel/plugin-proposal-private-methods', { loose: true }],
    ['@babel/plugin-proposal-private-property-in-object', { loose: true }],
    ['@babel/plugin-proposal-class-properties', { loose: true }],
    ['@babel/plugin-proposal-decorators', { legacy: true }],
    [
      '@babel/plugin-transform-runtime',
      {
        helpers: true,
        regenerator: true,
      },
    ],
]
```

amhed:
Inspecting transpiled JS it looks like decorator metadata is being issued: 

```js
__decorate([
  (0,decorators.field)('walletId'),
  __metadata("design:type", String)
[...]

PR #1857, opened 2024-11-21
Add instructions for vite

feznyng:
Adding some Vite web installation instructions (that worked for me) to address issues with field decorators (#1282).


radex:
Nice, thanks!

heliocosta-dev:
This was useful for me but I don't see it in the official docs website for some reason?

PR #1859, opened 2024-11-22
Electron support 

feznyng:
- Added remote adapter class which takes a callback used to actually execute queries. Implementers can provide a callback using any transport they want. I've added an example using electron's IPC to the docs. 
- This approach assumes arguments and response values passed to the adapter are serializable (Electron uses standard [structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)). Please let me know if this assumption is faulty. The alternative (and maybe better?) way is fully re-implementing the sqlite adapter but I'm trying to be lazy 😉.

Not really sure if this should be a draft or not, tests pass but I haven't added any extra for the new code. I won't have much time to work on this PR further but at the very least, might be helpful as reference for anyone trying to implement something similar.

### DIFF
--- a/docs-website/docs/docs/Installation.mdx
+++ b/docs-website/docs/docs/Installation.mdx
@@ -382,6 +382,26 @@ You only need this if you want to use WatermelonDB in NodeJS with SQLite (e.g. f
 
 ---
 
+## Electron (SQLite) setup
+
+You only need this if you want to use WatermelonDB in Electron with SQLite.
+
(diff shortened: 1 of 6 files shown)


radex:
Very cool work! Thank you @feznyng

LovesWorking:
Thank you!!

stiles3:
@LovesWorking, can this be accessed now? I would like to implement this electron feature in my app

maxwhipw:
If confirmed working can this pull request be accepted? @radex

PR #1879, opened 2025-01-20
Update sqlite makeDispatcher to show correct method name in error

gmacmaster:
Currently, the error shows `Cannot run database method undefined because database failed to open. Hint: Did you install JSI correctly? ...` when a user tries to call an unsupported method. 

Updated to show the method name the user is trying to call.


radex:
Thanks!

PR #1922, opened 2025-06-10
fix: make catchError visible to typescript

itsramiel:
`catchError` is very useful when finding and observing a model and that model possibly doesnt exist which then throws an error in that case `catchError` is used to handle the error. It is already being exposed and can be used but Typescript doesnt see the type definition.

I just exported it from the .d.ts files like the other operators


itsramiel:
@radex That's a super small one if you can take a look 🙏

radex:
thanks!

PR #1935, opened 2025-07-11
Add ezypack logo

wglad:


wglad:
Thanks Radek for including ezypack in the WatermelonDB docs.

PR #1943, opened 2025-08-10
Update Installation.mdx

mayank-01-ms:
This import is causing an error because the module is not resolved. If we remove this import, the build is successful, and the DB works completely fine.


radex:
Thanks!

mayank-01-ms:
@radex Any idea why the Kotlin instructions are not visible here https://watermelondb.dev/docs/Installation

radex:
@mayank-01-ms docs on the website are updated on release, not on merge to master

mayank-01-ms:
image


I am talking about this Kotlin tab. It was already present, but no tabs can be seen on the docs. Hoping it to get released as this might be really helpful for Kotlin users.


image

@radex

PR #1955, opened 2025-12-17
Fix/tag not found

jdicami:
Fixing race condition where database connection might not have been initialized yet.


jdicami:
Closing — handled in our internal fork.

PR #1968, opened 2026-05-29
fix: escape regex metacharacters in Q.like (LokiJS adapter diverges from SQLite)

sravan27:
## Summary

`likeToRegexp` (used by the LokiJS/web adapter to evaluate `Q.like`/`Q.notLike`) builds a `RegExp` from the LIKE pattern but never escapes regex special characters — it only maps `%`→`.*` and `_`→`.`. SQL `LIKE` treats every character other than `%`/`_` as a literal, and the SQLite adapter passes the pattern straight to SQL `LIKE`, so the two adapters disagree:

- **Silent wrong results:** `Q.like('%.pdf')` matches `'docXpdf'` on LokiJS (`.` = any char) but only real `.pdf` strings on SQLite; `Q.like('a.b')` matches `'axb'`.
- **Crash:** `Q.like('a(b')` / `Q.like('%(foo)%')` throws `SyntaxError: Invalid regular expression` (unbalanced group) on the LokiJS/web adapter.

## Fix

Escape regex metacharacters before substituting the `%`/`_` wildcards. The `s`/`i` flags are unchanged, so dotall + case-insensitive behaviour is preserved.

## Tests

Added `src/utils/fp/likeToRegexp/test.js`: wildcards, dotall + case-insensitivity (no regression), literal metacharacters (matching the SQLite adapter), and the no-crash cases. The mapping was also cross-checked against SQLite `LIKE` with a reference script — the current implementation diverges or crashes on 6/12 inputs; the fixed one matches all 12.

Opened as a draft pending a CI run.


MrCordeiro:
hi @radex , this looks like a good, minimal fix. Can it be merged?

codeCraft-Ritik (commented):
"Since this changes matching behavior for any existing app relying on the old (buggy) LokiJS behavior, should this go in as a semver-minor/patch with a CHANGELOG entry calling out the behavior change explicitly, in case anyone depended on the old loose matching?"

